PR #4, opened 2017-10-07
fix to documentation (correct endianess comment)

LivingInSyn:
both said little endian, changed the big endian one to the right comment


newpavlov:
Thank you for the fix!

PR #7, opened 2018-02-15
Add len_padding_with to specify the prefix byte

jmcomets:
This is needed in some variants where the prefix of the length padding is a
`0x01` byte, for example in [Tiger Hash](https://en.wikipedia.org/wiki/Tiger_(cryptography)).

I left `0x80` as the default padding byte for backwards compatibility.

Note that this is required for RustCrypto/hashes#50.

### DIFF
--- a/block-buffer/src/lib.rs
+++ b/block-buffer/src/lib.rs
@@ -69,10 +69,10 @@ macro_rules! impl_buffer {
             }
 
             #[inline]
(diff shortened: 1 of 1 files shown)


newpavlov:
Thank you! I will merge it now.

Although, I'll publish it in the following days as `v0.4` together with with dependence on the separate `block-padding` crate instead of internal `paddings` module, so I think I'll change `len_padding(_u128)` methods to include the prefix argument.

PR #10, opened 2018-10-03
Use while let

luisbg:
Loop is more readable as a 'while let'.

Thanks for creating this crypto related projects. Very interesting :)


newpavlov:
Thanks!

PR #11, opened 2018-10-03
fix substraction overflow

andy128k:
Currently bad input causes panic.


newpavlov:
Good catch! Thank you!

I will publish an update immediately.

PR #12, opened 2018-12-16
docs: Add usage for zero block padding

ltfschoen:


newpavlov:
Hm, zero padding and block modes which transform block cipher into stream cipher are not related to each other, so I am not sure if this addition is relevant.

PR #13, opened 2019-01-26
Migrate hex-literal to the 2018 edition

DarkEld3r:
Currently it is required to import `hex_literal` (in the 2018 edition crates) the following way:
```rust
use hex_literal::{hex, hex_impl};
```
After this update only `hex_literal::hex` is required.

I don't have enough experience with procedural macro error handling, but I suppose that previous errors reporting wasn't better.


newpavlov:
Thank you for your contribution!

I think I'll wait a bit before merging this PR and publishing v0.2, I plan to migrate most of RustCrypto crates to 2018 edition synchronously. Also maybe additionally we should migrate to true procedural macros instead of using `proc-macro-hack`.

PR #14, opened 2019-01-29
fix the buffer at the beginning of the struct

kazcw:
This allows the owner of a BlockBuffer to control the alignment of its
data, e.g. to use `movaps` instead of `movups`.

### DIFF
--- a/block-buffer/Cargo.toml
+++ b/block-buffer/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "block-buffer"
-version = "0.7.2"
+version = "0.7.3"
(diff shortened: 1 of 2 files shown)


newpavlov:
Hm, I am on the fence with this PR. Should we really use `repr(C)` for optimization purposes and make internal representation a part of backward compatibility promise? @tarcieri what do you think?

tarcieri:
Wonder if anyone working on const generics has thought about using them to control alignment

kazcw:
I was thinking about generics and I found a better way than this PR. See: #15.

PR #15, opened 2019-01-29
allow control of buffer alignment

kazcw:
Create a new API, AlignedBuffer, that is parameterized by a type used only for its alignment requirement. The BlockBuffer type becomes an alias for a u8-aligned AlignedBuffer.

I think it would be better to pass an &AlignAs<GenericArray> to the user callback to document the guarantees offered, but that would either be a breaking change or require a wrapper, and i'm not sure either is worth it: as it is, this PR adds alignment guarantees while remaining a drop-in replacement for the u8-aligned original.

### DIFF
--- a/block-buffer/Cargo.toml
+++ b/block-buffer/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "block-buffer"
-version = "0.7.2"
+version = "0.7.3"
 authors = ["RustCrypto Developers"]
 license = "MIT OR Apache-2.0"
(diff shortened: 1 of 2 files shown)


kazcw:
Oh, this doesn't work because input() bypasses the buffer to stream directly from unaligned data. Avoiding the copy when hashing long inputs is more important than avoiding the `movups` when using the buffer, and doing both would complicate the interface too much. So I think aligned buffers in general are a no-go for this API.

PR #16, opened 2019-02-22
Removes unnecessary backtick from hex-literal doc

plafue:
The extra backtick breaks formatting in docs.rs


newpavlov:
Thank you!

Issue #20, opened 2019-03-14
Padding requires too much overhead


sebastianv89:
The first encryption example on the README page is somewhat counter-intuitive: the 12-byte message `b"Hello world!" is encrypted to 16 bytes, but the buffer must be 24 bytes (it is actually 32 bytes in the example).

It turns out that none of the bytes `buffer[16..]` are ever read or even used for intermediate computations. Instead, the `block_padding::Padding` trait has a default method `pad` that requires `buf.len() - pos >= block_size`. This can result in a `PadError` despite there being enough space for the padding.

newpavlov:
I've published `block-padding` v0.1.4 with your update!

PR #21, opened 2019-03-14
Relaxed restrictions on padding buffer size.  See #20

sebastianv89:


block-padding/src/lib.rs
```diff
@@ -31,18 +31,21 @@ pub trait Padding {
 
     /// Pads message with length `pos` in the provided buffer.
     ///
-    /// `&buf[..pos]` is percieved as the message, buffer must contain at
-    /// least one block of leftover space, i.e. `buf.len() - pos >= block_size`
-    /// must be true. Otherwise method will return `PadError`.
+    /// `&buf[..pos]` is perceived as the message, the buffer must contain
+    /// enough leftover space for padding: if `pos % block_size == 0` then
+    /// a full extra block must be available, else `block_size - (pos %
+    /// block_size)` extra bytes must be available. Otherwise method will
```
newpavlov: Aren't those two conditions fully equivalent to each other? You can change code accordingly.

sebastianv89: Yes, of course! I updated the code

PR #23, opened 2019-04-15
Make block-buffer minimal-versions correct

CAD97:
byteorder:1.0.0 does not export byteorder::BE.
It was added in byteorder:1.1.0.

Noticed in https://github.com/pest-parser/pest/pull/390.


newpavlov:
Oh! Thank you! I've missed implementation of `minimal-version`. I will need to add it to CI.

PR #24, opened 2019-04-15
Fix minimal version of byteorder.

dragostis:
block-buffer fails to build with cargo -Z minimal-versions since it's relying
on an [alias] from byteorder that was made available only in 1.1.

[alias]: https://github.com/RustCrypto/utils/blob/master/block-buffer/src/lib.rs#L7


dragostis:
Closed in favor of #23.

PR #28, opened 2019-07-30
Don't depend on user crate's std feature

xfix:
This avoids a compiler error when used with edition 2015 std crates that don't have an std feature.


newpavlov:
Thank you! Looks good!

PR #30, opened 2019-08-23
add no_std to hex-literal

camshaft:
When including this library in a crate with `#![no_std]` forces the std library to be included. Thanks!

### DIFF
--- a/hex-literal/src/lib.rs
+++ b/hex-literal/src/lib.rs
@@ -27,6 +27,7 @@
 //! ```
 #![doc(html_logo_url =
     "https://raw.githubusercontent.com/RustCrypto/meta/master/logo_small.png")]
+#![no_std]
 
 use proc_macro_hack::proc_macro_hack;
 #[doc(hidden)]


Recmo:
Ha! I was just about to submit a similar PR. I have this feature flagged:

```toml
[features]
default = []
std = []
```
```rust
#![cfg_attr(not(feature = "std"), no_std)]
```
Although I'm not sure if this is necessary. Is mixin `no_std` in `std` code always without downsides?

camshaft:
as far as I'm aware you're free to use `#![no_std]` in any crates without downsides. All of the other utility crates in this workspace [do the same](https://github.com/RustCrypto/utils/blob/master/digest-buffer/src/lib.rs#L1); it just looks like `hex-literal` was missed.

newpavlov:
Thank you! I didn't add `#![no_std]` previously because `hex-literal-impl` uses `Vec`, so I was not sure if it will make any difference. I will release v0.2.1 right away.

PR #35, opened 2019-10-31
Remove unsafe{} code from the byte-tools crate

DarkKirb:
While the current unsafe code is sound, it is possible to express `byte-tools`'s functions without using unsafe blocks.

This change should not have a significant performance impact in release mode, but the `zero` and `set` functions will be slower in debug mode.


newpavlov:
This change will add a needless [branching](https://rust.godbolt.org/z/hwsBz6). While it can be eliminated after inlining and correct branch is easily predictable, I don't think it's worth to remove `unsafe` in this particular case.

BTW in future I plan to significantly reduce usage of `byte-tools` crate, so it may dissapear from your dependency tree eventually.

PR #36, opened 2019-11-05
feat(block-buffer): add input2 back

dignifiedquire:
This implementation now works on rust@1.21.0 as well.

It simply copies `std::from_slice` inline from here: https://doc.rust-lang.org/src/core/slice/mod.rs.html#5313-5317


newpavlov:
I will close it in favor of #56.

Issue #34, opened 2019-10-23
block-padding does not provide a "no padding" implementation


robeady:
I found it rather puzzling that this library does not provide a no-op / no padding implementation, requiring the user to do it themselves. I expected to find one since rust-crypto does provide this.

zer0x64:
+1, had the exact same problem trying to implement a protocol I didn't write. Also, some operation like key wrapping might actually require a NoPadding option for legit purposes. Right now, the "best available option" is to use ZeroPadding and check it the size is a mutliple of the block size to re-pad it in case the value actually ends with zeroes.

newpavlov:
@robeady 
Does #37 address your needs?

robeady:
I believe it does, thanks :)

The problem with ZeroPadding is that it would remove trailing zeros from blocks when unpadding. I initially assumed that zero padding meant no padding which caused some confusion lol

newpavlov:
>The problem with ZeroPadding is that it would remove trailing zeros from blocks when unpadding.

Ah, yes. You are right.

PR #37, opened 2019-11-06
NoPadding for block-padding

zer0x64:
Added a NoPadding option for block-padding crate. Relates to https://github.com/RustCrypto/utils/issues/34

Note that I had to reimplement the `pad()` function for this one since the original function created a new block when `data_size % block_size == 0`, which is not intended for this mode. This makes the pad_block function useless, but I kept it instead of using an `unreachable!()` since the function is publicly exposed, giving it the same implementation as `pad`.

### DIFF
--- a/block-padding/src/lib.rs
+++ b/block-padding/src/lib.rs
@@ -275,3 +275,49 @@ impl Padding for Iso7816 {
         Ok(&data[..n])
     }
 }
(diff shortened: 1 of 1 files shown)


newpavlov:
Thank you!

I will do a release with this update right away.

PR #53, opened 2020-06-01
hex-literal v0.3

### LINKED ISSUE #25: Remove dependency on proc-macro-hack after proc_macro_expr stabilization
Right now we have to use `proc-macro-hack` in `hex-literal` to circumvent the following error: "procedural macros cannot be expanded to expressions (see issue rust-lang/rust#54727)". After use of procedural macros will be allowed in expressions we will be able to use `proc-macro` directly to implement the required functionality.

newpavlov:
This version will work only in Rust 1.45 and later.

Closes #25.


hex-literal/Cargo.toml
```diff
@@ -9,6 +9,5 @@ repository = "https://github.com/RustCrypto/utils"
 keywords = ["hex", "proc-macro", "literals"]
 edition = "2018"
 
-[dependencies]
-proc-macro-hack = "0.5"
```
tarcieri, line 13: Hooray for `proc_macro_hygiene`! 😄

newpavlov:
Hm, weird. CI runs for `stable` pass without issues, but `1.45.0` fails... In the log we can see:
```
info: installing component 'rustc'

  1.45.0-x86_64-unknown-linux-gnu installed - rustc 1.45.0 (5c1f21c3b 2020-07-13)

info: checking for self-updates
Gathering installed versions
  /usr/share/rust/.cargo/bin/rustc -V
  rustc 1.44.1 (c7087fe00 2020-06-17)
  /usr/share/rust/.cargo/bin/cargo -V
  cargo 1.44.1 (88ba85757 2020-06-11)
  /usr/share/rust/.cargo/bin/rustup -V
  rustup 1.22.1 (b01adbbc3 2020-07-08)
```

tarcieri:
@newpavlov it's getting installed, however I think you need to add `override: true` here:

https://github.com/RustCrypto/utils/blob/master/.github/workflows/hex-literal.yml#L51

It looks like both of them start out with `1.44.1` as `stable`. When you upgrade `stable` it's working (because that's the default toolchain), but without `override: true` it's installing 1.45.0 but not using it because the default toolchain is still set to `stable` (which is still `1.44.1`)

newpavlov:
Ah, indeed. We probably will need to check other repositories, since the config was based on a copy (although I don't remember from where).

PR #62, opened 2020-06-11
Add cpuid-bool crate

newpavlov:
A specialized alternative to `is_x86_feature_detected`. This crate is no-std friendly and caches `bool`, instead of all CPUID flags. I plan to use it for SHA-1 and SHA-2 crates, but it could be usefull for other crates as well, e.g. for `chacha20`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,6 +5,7 @@ members = [
     "block-padding",
     "byte-tools",
     "collectable",
+    "cpuid-bool",
     "dbl",
     "fake-simd",
     "hex-literal",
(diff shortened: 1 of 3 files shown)


newpavlov:
Well, as I've said in the issue, I am a bit hesitant about adding runtime detection for AES on block-cipher level. With hashes and ChaCha20 we can enjoy the fact that SIMD implementations have the same memory layout as software fallbacks, which is drastically different for AES. At the very least I hope we will be able to reduce memory footprint of the bitsliced implementation.

tarcieri:
Aah yes, I suppose a better bitsliced implementation is in order first.

PR #64, opened 2020-06-17
blobby: new storage format

newpavlov:
The new format uses git-flavored [VLQ](https://en.wikipedia.org/wiki/Variable-length_quantity) to encode length of binary blobs. Additionally it supports referencing of previously encountered values, which should be especially helpful for de-duplication of keys and plaintexts.

Since the new format is incompatible with the old one, we would have to re-encode existing `blb` files, which probably will have to wait until the next minor releases.


tarcieri:
I’m reminded of https://github.com/iqlusioninc/veriform/tree/develop/rust/vint64/ 😉

newpavlov:
Note, that it's a tiny bit less space efficient, so technically the third claim is false. :) Max integer for 2 byte git-VLQ is 16512, while in your crates it's 16383. But on the other side, it probably uses less branches, so it should be a bit more efficient at runtime.

blobby/src/lib.rs
```diff
@@ -1,206 +1,248 @@
...
-    pub fn new(data: &'a [u8]) -> Result<Self, &'static str> {
-        if data.len() < 7 { Err("data is too small")? }
-        let (header, data) = data.split_at(7);
-        let size = match header {
-            b"blobby1" => IndexSize::N8,
-            b"blobby2" => IndexSize::N16,
-            b"blobby4" => IndexSize::N32,
-            b"blobby8" => IndexSize::N64,
-            _ => Err("invalid data header")?,
+    idx: [&'a [u8]; INDEX_SIZE],
```
newpavlov, line 44: I don't quite like this part, but I see only two alternatives:
- Sacrifice no-heap property and store this index on the heap.
- Traverse the index blobs each time, which is obviously much worse performance wise.

@tarcieri 
What do you think?

tarcieri: I'm not sure I understand the format or the problem well enough, but I can not that a well-designed format can be decoded in a single-pass without the need for the heap (which would seem to defeat the point of blobby in the first place?)

newpavlov: So initially I was using simple offsets, which can be easily processed in a single pass. But offsets usually bigger than 63 bytes, so they will not fit into one byte. This is why I tried to use an index number of a de-duplicated blob instead. Most often encountered blobs are stored near the beginning, so most reference entries fit nicely into one byte. But it causes a problem: we must store mapping "serial number" -> "position of a blob" somewhere. In other words here I traded easiness of parsing for space minimization.

The current code will waste a certain amount of stack space, but I wonder if it will be a big problem, considering the main use-case of this format.

newpavlov: I've changed the code to boxed slice. While dependency on `alloc` is not ideal, it looks like it should not be a big obstacle for most embedded targets. Especially considering that AEAD tests right now either way are dependent on `alloc`.

PR #66, opened 2020-07-02
Remove deprecated crates

newpavlov:
We probably should create a separate repository to track deprecated crates.


newpavlov:
Note that `hex-literal` and `opaque-debug` will work on Rust 1.31, but the new Cargo.lock format is properly understood only by Rust 1.41 and later.

PR #68, opened 2020-07-14
cpuid-bool: fix SGX support (take 2)

newpavlov:
Fixed version of #67


tarcieri:
Does this mean all the feature flags will be false inside an SGX environment?

newpavlov:
Not quite, `cpuid-bool` [takes into account ](https://github.com/RustCrypto/utils/blob/master/cpuid-bool/src/lib.rs#L114-L117) target features enabled via compiler flags.

tarcieri:
Aah... well then

PR #76, opened 2020-08-04
const-oid validation

tarcieri:
~~Attempting to use the new `const fn` features set to ship in 1.46, however it seems this may not be a good idea until const panic support ships and malformed OIDs can panic at compile time~~

Uses a workaround proposed by @rodrimati1992 below to allow a limited degree of compile-time const OID validation when used in a `const` context (which is what this crate is designed for, after all).

### DIFF
--- a/.github/workflows/const-oid.yml
+++ b/.github/workflows/const-oid.yml
@@ -22,7 +22,7 @@ jobs:
     strategy:
       matrix:
         rust:
-          - 1.39.0 # MSRV
+          - 1.46.0 # MSRV
           - stable
         target:
           - thumbv7em-none-eabi
@@ -41,7 +41,7 @@ jobs:
     strategy:
       matrix:
         rust:
-          - 1.39.0 # MSRV
+          - 1.46.0 # MSRV
           - stable
     steps:
       - uses: actions/checkout@v1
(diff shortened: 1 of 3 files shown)


rodrimati1992:
If this is mostly intended to be called at compile-time, you can hack a panic-with-error-message using a single-element array like:
https://play.rust-lang.org/?version=beta&mode=debug&edition=2018&gist=2c42d011572caab4fd3f7704c6e92e90
```rust
error: any use of this value will cause an error
  --> src/lib.rs:15:9
   |
15 |         ["invalid OID"][is_invalid as usize];
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |         |
   |         index out of bounds: the len is 1 but the index is 1
   |         inside `ObjectIdentifier::new` at src/lib.rs:15:9
   |         inside `OID` at src/lib.rs:21:35
...
21 | pub const OID: ObjectIdentifier = ObjectIdentifier::new(&[0, 60, 3]);
   | ---------------------------------------------------------------------
   |
   = note: `#[deny(const_err)]` on by default
```

tarcieri:
Hax! Thanks, good to know.

PR #83, opened 2020-10-02
Biguint-literal crate

newpavlov:
Drafted for RustCrypto/elliptic-curves#218.


newpavlov:
I still need this branch.

tarcieri:
Err whoops, not sure what happened here. Sorry about that.

tarcieri:
@newpavlov I think `crypto-bigint` might accomplish the same sort of goal as what's in this PR, but using `const fn` instead of a proc macro. Here's an example:

https://github.com/RustCrypto/traits/pull/645/files#diff-369e56ff85c2cd28f4a387c6a341ae8c5880633480cb4f629b30f9c8a79523c5R45-R46

newpavlov:
Yes, I've noticed that you've managed to implement `const` initialization and was pleasantly surprised that it is possible with the current version of const generics. I will close this PR after I migrate RustCrypto/elliptic-curves#218 to `crypto-bigint`. But I will have to start with learning more about how `crypto-bigint` is implemented, since I only have been able to take a cursory look at it before.

tarcieri:
FYI, I'll probably look at adding some basic support for field elements to `crypto-bigint`, although unfortunately const generics aren't quite powerful enough yet to model them the way I'd really like (i.e. with the field modulus expressed as a const generic)

They'll probably not be ideal for things like elements of the base field, but they should be fine for representing scalar fields.

PR #86, opened 2020-11-28
Improve cpuid-bool efficiency

newpavlov:
The new API creates a module which contains `InitToken` zero sized type. The only way to create a value of this type is via functions which ensure that initialization code has been executed. It should be slightly more efficient for our use-cases, since initialization branch can be eliminated from methods which work with initialized algorithm (e.g. `update` and `encrypt_blocks/decrypt_blocks`).

In future it could be beneficial to add "cpuid enums", i.e. macro which would generate code not for a `bool`, but for a user-specified enum.

### DIFF
--- a/cpuid-bool/src/lib.rs
+++ b/cpuid-bool/src/lib.rs
@@ -2,16 +2,31 @@
 //!
 //! # Usage example
 //! ```
-//! if cpuid_bool::cpuid_bool!("sha", "aes") {
+//! cpuid_bool::new!(cpuid_aes_sha, "aes", "sha");
+//!
+//! // `token` is a Zero Sized Type value, which guarantees
+//! // that underlying static storage got properly initialized,
+//! // which allows to omit initialization branch
+//! let token = cpuid_aes_sha::init();
+//! if token.get() {
 //!     println!("CPU supports both SHA and AES extensions");
 //! } else {
 //!     println!("SHA and AES extensions are not supported");
 //! }
+//! 
+//! // If stored value needed only once you can get stored value
+//! // omitting the token
+//! let val = cpuid_aes_sha::get();
+//! assert_eq!(val, token.get());
+//!
+//! // Additionally you can get both token and value
+//! let (token, val) = cpuid_aes_sha::init_get();
+//! assert_eq!(val, token.get());
 //! ```
 //! Note that if all tested target features are enabled via compiler options
-//! (e.g. by using `RUSTFLAGS`), `cpuid_bool!` macro immideatly will expand
-//! to `true` and will not use CPUID instruction. Such behavior allows
-//! compiler to eliminate fallback code.
+//! (e.g. by using `RUSTFLAGS`), the `get` method will always return `true`
+//! and `init` will not use CPUID instruction. Such behavior allows
+//! compiler to completely eliminate fallback code.
 //!
 //! After first call macro caches result and returns it in subsequent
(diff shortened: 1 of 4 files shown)


newpavlov:
Clippy errors are unrelated to changes in this PR.

cpuid-bool/src/lib.rs
```diff
@@ -1,51 +1,118 @@
...
+//! let (token, val) = cpuid_aes_sha::init_get();
+//! assert_eq!(val, token.get());
 //! ```
 //! Note that if all tested target features are enabled via compiler options
-//! (e.g. by using `RUSTFLAGS`), `cpuid_bool!` macro immideatly will expand
-//! to `true` and will not use CPUID instruction. Such behavior allows
-//! compiler to eliminate fallback code.
+//! (e.g. by using `RUSTFLAGS`), the `get` method will always return `true`
+//! and `init` will not use CPUID instruction. Such behavior allows
+//! compiler to completely eliminate fallback code.
```
tarcieri, line 30: I was a bit confused by this section (which I guess was there to begin with). It would seem to defeat the point of having runtime detection at all?

One of the main use cases I have in mind is when compiled binaries are relocated to environments where one or more of the features are unavailable, e.g. a VM which does not support AES-NI.

newpavlov: Programs compiled with enabled target features are not fully portable in principle. If you enable non-default target features (e.g. AVX2 or AVX512), compiler in its full right to use those instructions to optimize code. Thus the resulting binary would not work in environments which do not support the enabled extension. If you can not guarantee that program will not be executed only on CPUs with AES-NI support you should not enable the `aes` target feature via compiler flags but if you can not guarantee it, then by enabling it you will get a bit more efficient binary (smaller size of the AES types and lack of fallback branches).

I am not happy with the current description, so I will gladly replace it with a better explanation if you have one.

tarcieri: To break the problem down as I understand it:

- In order for Rust to emit AES-NI instructions, the `target_feature` for `aes` *MUST* be enabled. If the target feature is not enabled, LLVM will use a software implementation of the AES-NI instructions (see [discussion here](https://github.com/RustCrypto/block-ciphers/pull/208#issuecomment-735400697))
- Even on the same *physical* CPU, it's possible for AES-NI to be enabled or disabled, depending on whether AES-NI is enabled or disabled in the BIOS, or whether the workload is virtualized and virtualization supports or does not support AES-NI. This is a bit different from other CPU instructions which we could otherwise assume are always available given some CPU microarchitecture as a baseline.

So I think the approach in this PR can work, but we'll need e.g. the `aes` target feature enabled at the time the program is compiled, and for `cpuid_bool` to always perform the check regardless of enabled target features.

tarcieri (approved):
Okay sorry, it took me awhile to figure out this is all working as expected, but now I've tested it out empirically and it seems to be working the way you described.

cpuid-bool/src/lib.rs
tarcieri: Never mind, I was wrong.

tarcieri:
Fixed clippy in #87 and rebased. Should be good to merge/release now.

tarcieri:
@newpavlov can you publish the crate? I don't have access.

newpavlov:
I plan to do it on the 1st of December as specified in the changelog. I will add you to the owners list a bit later.

PR #89, opened 2020-12-04
const-oid: byte and string parsers

tarcieri:
This commit adds backing storage to `ObjectIdentifier` and with it the ability to parse them at runtime from BER/DER bytestrings and dot-delimited strings.

### DIFF
--- a/const-oid/src/lib.rs
+++ b/const-oid/src/lib.rs
@@ -6,6 +6,10 @@
 //! It has full `no_std` support with no dependencies on a heap/liballoc and
 //! stores OID values as static data.
 //!
+//! The [`ObjectIdentifier`] type can also be used for runtime modeling of OIDs
+//! (e.g. when parsing messages), with the goal of easily comparing them to
+//! constant values.
+//!
 //! # About OIDs
 //!
 //! Object Identifiers, a.k.a. OIDs, are an International Telecommunications Union (ITU) and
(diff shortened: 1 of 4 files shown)


const-oid/src/lib.rs
```diff
@@ -21,50 +25,183 @@
...
+            ],
+            9 => [
+                nodes[0], nodes[1], nodes[2], nodes[3], nodes[4], nodes[5], nodes[6], nodes[7],
+                nodes[8], 0,
+            ],
+            10 => [
+                nodes[0], nodes[1], nodes[2], nodes[3], nodes[4], nodes[5], nodes[6], nodes[7],
+                nodes[8], nodes[9],
+            ],
+            _ => [0u32; MAX_NODES], // Checks above prevent this case, but makes the compiler happy
```
newpavlov, line 130: Use `unreachable!()` here?

const-oid/src/lib.rs
```diff
@@ -21,50 +25,183 @@
...
+
+/// Maximum value of the first node in an OID
+const FIRST_NODE_MAX: u32 = 2;
+
+/// Maximum value of the second node in an OID
+const SECOND_NODE_MAX: u32 = 39;
+
+/// Error type
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+pub struct Error;
```
newpavlov, line 51: Implement the `std::error::Error` trait?

newpavlov (commented):
I haven't thoroughly checked the BER/DER decoding, but otherwise looks good.

const-oid/src/lib.rs
tarcieri: It's a `const fn`, so panic is unsupported.

Best I could do is `unreachable_unchecked()`, but that seems like a needless use of `unsafe` in an otherwise `forbid(unsafe_code)` crate.

const-oid/src/lib.rs
tarcieri: Yeah, I can do that. So annoying to have to add a `std` feature for that though.

PR #94, opened 2020-12-04
pkcs8: length handling improvements

tarcieri:
Rejects some cases that were previously accepted but MUST be rejected by ASN.1 DER when parsing lengths:

- When the length is between 0 and 127, the short form of length must be used
- Ensures a minimum number of octets are used by disallowing 3-byte lengths for values less than 256 (per X.690 Section 10.1)
- Rejects a leading length byte of 128 (0x80), as it is reserved for indefinite lengths (per X.690 Section 8.1.3.6.1)

Note: per RFC 5208, PKCS#8 is technically BER encoded (despite what things like the OpenSSL's `-outform DER` would lead you to believe).

Despite that, this PR attempts to accept only valid DER, as that's how PKCS#8 keys are typically described and hopefully, in practice, most PKCS#8 keys *are* actually valid DER. We can revisit this if it turns out we're rejecting valid BER-encoded PKCS#8 keys.

### DIFF
--- a/pkcs8/src/asn1.rs
+++ b/pkcs8/src/asn1.rs
@@ -1,4 +1,12 @@
-//! PKCS#8 parser
+//! PKCS#8 parser: supports the subset of ASN.1 DER needed to parse PKCS#8.
+//!
+//! Note: per RFC 5208, PKCS#8 is technically BER encoded (despite what
+//! things like the OpenSSL's `-outform DER` would lead you to believe).
+//!
+//! Despite that, this implementation presently attempts to reject BER-encoded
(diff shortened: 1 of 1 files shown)


tarcieri:
I'm a bit torn on this PR, especially as it's something of a half-measure.

It seems like it would be better to either go fully DER, even if RFC 5208 doesn't require it, or completely abandon DER and except any valid BER-encoded document, and further rename all current references to DER to BER.

I'm worried that given how ubiquitously PKCS#8 documents seem to be described as DER, however, that it might be confusing to use BER instead, even if it's what the RFC stipulates.

PR #113, opened 2020-12-06
block-buffer and block-padding improvements

### LINKED ISSUE #79: block-padding: add NAME associated constant to the Padding trait
Since padding types are not populated we can not use derived `Debug` impl, so for opaque debug implementations it would be useful to have an associated `&'static str` name constant. This is a breaking change.

newpavlov:
This PR modifies `block-buffer` and `block-padding` APIs to reduce their surface and to make them mostly panic-free. Lack of panics in `block-buffer` was checked using godbolt. The relevant `unsafe` code line relies on the following invariant: `pos` is always strictly smaller than block size.

Because of this invariant I had to remove the `input_lazy` method, which was initially added for Skein. I think it will be better to later introduce a separate "lazy" buffer type. As a consequence it means that Skein will not be compatible with the core API introduced in RustCrypto/traits#380 (I think it could be eventually fixed with specialization). For most users it will be a barely noticeable implementation detail (i.e. Skein will simply implement the mid-level traits directly, without relying on the core wrapper), so considering relative unpopularity of hashes reliant on "lazy" buffers, I think it's a reasonable sacrifice.

Ideally we would use an Ada-like range type for `pos` (both the `block-buffer` field and the `block-padding` method argument), but without const generics such type probably will be quite unergonomic.

Closes #79

### DIFF
--- a/block-buffer/src/lib.rs
+++ b/block-buffer/src/lib.rs
@@ -15,7 +15,9 @@ use block_padding::Padding;
 use core::{convert::TryInto, slice};
 use generic_array::{typenum::U1, ArrayLength, GenericArray};
 
+/// Block on which a `BlockBuffer` operates.
 pub type Block<BlockSize> = GenericArray<u8, BlockSize>;
+/// Blocks being acted over in parallel.
 pub type ParBlock<BlockSize, ParBlocks> = GenericArray<Block<BlockSize>, ParBlocks>;
 
 /// Buffer for block processing of data.
--- a/block-buffer/src/lib.rs
+++ b/block-buffer/src/lib.rs
@@ -15,10 +15,13 @@ use block_padding::Padding;
 use core::{convert::TryInto, slice};
 use generic_array::{typenum::U1, ArrayLength, GenericArray};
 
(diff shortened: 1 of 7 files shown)


newpavlov:
I have added the `xor_data` method, which should be useful for RustCrypto/stream-ciphers#95. I am not 100% sure it will be universal enough to unify buffering logic for one-pass AE and AEAD implementations, but it should be a good enough starting point.

block-buffer/src/lib.rs
```diff
@@ -1,241 +1,274 @@
...
+        self.set_pos(n);
     }
 
-    /// Variant that doesn't flush the buffer until there's additional
-    /// data to be processed. Suitable for tweakable block ciphers
-    /// like Threefish that need to know whether a block is the *last*
-    /// data block before processing it.
+    /// XORs `data` using provided functions.
+    ///
+    /// This methof is intended for stream cipher implementations. If `N` is
```
tarcieri, line 95:
```suggestion
    /// This method is intended for stream cipher implementations. If `N` is
```

tarcieri (approved):

PR #188, opened 2020-12-26
WIP: block-buffer: drop generic-array for min_const_generics

linkmauve:
This bumps the minimum required Rust version.

This shouldn’t be merged until `min_const_generics` gets stabilised, but once it is it will allow to remove the generic-array crate from this one.

### DIFF
--- a/block-buffer/Cargo.toml
+++ b/block-buffer/Cargo.toml
@@ -12,4 +12,3 @@ edition = "2018"
 
 [dependencies]
 block-padding = { version = "0.2.0", path = "../block-padding", optional = true }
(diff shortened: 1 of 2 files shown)


newpavlov:
I think it's too soon for migrating to const generics for us. `min_const_generics` are not enough to cover needs of this organization, the main blocker is rust-lang/rust#60551. We probably will do a coordinated update of all crates once stable const generics will be sufficiently advanced.

tarcieri:
Oh wow, https://github.com/rust-lang/rust/issues/60551 is quite the showstopper for our usage.

PR #200, opened 2021-01-12
b64ct: initial crate

tarcieri:
Extraction of the "B64" implementation presently located in the `password-hash` crate:

https://github.com/RustCrypto/traits/blob/dda4f9b/password-hash/src/b64.rs

Originally based on the following C++ implementation:

https://github.com/Sc00bz/ConstTimeEncoding/blob/master/base64.cpp

Original C++ is MIT licensed. Relicensed as Apache 2.0+MIT with permission from the original author.


tarcieri:
Note: the goal of extracting this into its own crate is to use it for PEM encoding/decoding in the `pkcs8` crate.

PR #204, opened 2021-01-12
Rustified panic-free modification of b64ct

newpavlov:
This PR adds two `unsafe` line for conversion into strings. We could remove them by using byte slices/vectors instead of strings. I am not sure how useful in practice is to have `&str`/`String` instead of `&[u8]`/`Vec<u8>`.

Panic-free property was verified in [godbolt](https://rust.godbolt.org/z/qe5Y5z).

Errors become a bit verbose, ideally it would be nice to have "anonymous enums" (i.e. `InvalidLengthError | InvalidEncodingError`), but we have to work with what we have.

One potential optimization is to modify `decode` API to:
```rust
// `buf` contains a B64-encoded string
pub fn decode<'a>(buf: &'a mut [u8]) -> Result<&'a [u8], InvalidEncodingError> { .. }
```
This way we will be able to reuse the buffer and remove the `Error` enum, but, again, I am not sure how such change would influence downstream crates.

### DIFF
--- a/b64ct/src/errors.rs
+++ b/b64ct/src/errors.rs
@@ -1,0 +1,63 @@
+use core::fmt;
+
+/// Insufficient output buffer length.
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+pub struct InvalidLengthError;
+
+impl fmt::Display for InvalidLengthError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
+        f.write_str("insufficient output buffer length")
+    }
+}
+
+#[cfg(feature = "std")]
+impl std::error::Error for InvalidLengthError {}
+
+/// Invalid encoding of provided "B64" string.
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+pub struct InvalidEncodingError;
+
+impl fmt::Display for InvalidEncodingError {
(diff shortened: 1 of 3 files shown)


b64ct/src/errors.rs
```diff
@@ -0,0 +1,63 @@
...
+    InvalidEncoding,
+    /// Invalid encoding of provided "B64" string.
+    InvalidLength,
+}
+
+impl fmt::Display for Error {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
+        match self {
+            v @ Self::InvalidEncoding => v.fmt(f),
+            v @ Self::InvalidLength => v.fmt(f),
```
tarcieri, line 42: Looks like this is breaking the build due to infinite recursion. Would suggest something like:

```suggestion
            v @ Self::InvalidEncoding => f.write_str("invalid B64 encoding"),
            v @ Self::InvalidLength => f.write_str("invalid B64 length"),
```

newpavlov: Hm, weird... It was building fine on my PC.

tarcieri (approved):
Cool, looks good!

An in-place API sounds good, but I'd suggest doing it additively, which is to say in addition to the current APIs. In the cases where this crate is presently used the source slice is immutable, so switching to an in-place API would add an extra copy.

newpavlov:
I probably will try to add the inplace variant in a separate PR then. It's a bit non-trivial since we have to correctly work around aliasing issues.

PR #209, opened 2021-01-13
b64ct: modify decode_in_place to be panic-free

newpavlov:
Number of `unsafe` blocks in `decode_in_place` can be minimized to two. Panic-free property was verified using [godbolt](https://rust.godbolt.org/z/jTYjzq). To be extra safe I also run the tests via MIRI and it didn't find any issues. We probably should add MIRI runs to CI for every crate which contains `unsafe` code. Also it could be worth to run fuzzing after #205 will be fixed.

This modification also significantly improves performance of `decode_in_place` from 150 MB/s to 205 MB/s on my laptop, now it's slightly faster than `decode` which results in 200 MB/s.


tarcieri (approved):
Nice

PR #212, opened 2021-01-13
b64ct: fix overflow bug, add alloc-gated tests

### LINKED ISSUE #205: b64ct: subtle overflow bug

newpavlov:
Closes #205

### DIFF
--- a/b64ct/src/lib.rs
+++ b/b64ct/src/lib.rs
@@ -44,7 +44,10 @@ pub use errors::{Error, InvalidEncodingError, InvalidLengthError};
 /// Encode the input byte slice as "B64", writing the result into the provided
 /// destination slice, and returning an ASCII-encoded string value.
 pub fn encode<'a>(src: &[u8], dst: &'a mut [u8]) -> Result<&'a str, InvalidLengthError> {
-    let elen = encoded_len(src);
+    let elen = match encoded_len_inner(src.len()) {
+        Some(v) => v,
+        None => return Err(InvalidLengthError),
+    };
     if elen > dst.len() {
         return Err(InvalidLengthError);
     }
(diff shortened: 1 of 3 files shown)


b64ct/src/lib.rs
```diff
@@ -70,10 +73,13 @@ pub fn encode<'a>(src: &[u8], dst: &'a mut [u8]) -> Result<&'a str, InvalidLengt
...
 
 /// Encode the input byte slice as a "B64"-encoded [`String`].
+///
+/// # Panics
+/// If `input` length is greater than `usize::MAX/4`.
 #[cfg(feature = "alloc")]
 #[cfg_attr(docsrs, doc(cfg(feature = "alloc")))]
 pub fn encode_string(input: &[u8]) -> String {
-    let elen = encoded_len(input);
+    let elen = encoded_len_inner(input.len()).expect("input is too big");
```
tarcieri, line 82: I'd be ok with making this API fallible for this case

tarcieri (approved):
Great, thank you!

b64ct/src/lib.rs
newpavlov: I am not sure if it's worth the trouble. In the first place it's a convenience method and probability of encountering values larger than 1 GiB is extremely small. If someone wants to be strictly panic-free, they always can use the underlying function directly.

b64ct/src/lib.rs
```diff
@@ -200,7 +221,18 @@ pub const fn decoded_len(bytes: &str) -> usize {
...
+    type DoubleWord = u128;
+    #[cfg(target_pointer_width = "32")]
+    type DoubleWord = u64;
+    #[cfg(target_pointer_width = "16")]
+    type DoubleWord = u32;
+
+    if n <= usize::MAX / 3 {
+        (3 * n) / 4
+    } else {
+        ((3 * (n as DoubleWord)) / 4) as usize
```
newpavlov, line 234: I think it should be possible to do it in a branch-less fashion without introducing the `DoubleWord` type...

newpavlov:
Matches and if branches in `const fn`s have MSRV of 1.46. What do you think is worth doing here? Bump crate MSRV, make the functions non-const, try to find branch-less tricks?

tarcieri:
MSRV 1.46+ is fine by me!

All of the dependent crates which use `b64ct` (i.e. `password-hash` and `pkcs8`) are MSRV 1.46+ anyway.

Issue #220, opened 2021-01-22
b64ct -> base64ct?


tarcieri:
I originally created the `b64ct` crate as an extraction of the ["B64" (in the PHC string format sense)](https://github.com/P-H-C/phc-string-format/blob/master/phc-sf-spec.md#b64) code in the `password-hash` crate.

For context, "B64" is Base64 without padding (i.e. `=`) and whitespace, using the normal Base64 alphabet.

However, I've gone to use it for two other purposes relating to private key formats (which could really benefit from a constant-time implementation) which go beyond what is possible with the crate today:

- `pkcs8` is using it for its "PEM" (in the RFC 7468 sense) implementation, which requires chunking at 64 encoded characters (i.e. inserting newlines), and also padding with `=`
- `elliptic-curve` is now using it for its JWK implementation, which requires Base64uri. It's doing this by running a substitution across a `b64ct` string (which largely defeats the point of a constant-time implementation)

I think it'd be good to expand `b64ct` into a `base64ct` crate which supports at least these three flavors:

- "B64" for `password-hash`
- "PEM encoding" for PKCS#8
- Base64uri for JWT

newpavlov:
Sounds good! Though depending on amount of code reuse it may be worth to keep those flavors in separate crates, for example I am not sure how the current code will fit to the `pkcs8` use-case.

tarcieri:
There's not really much to it: https://github.com/RustCrypto/utils/blob/master/pkcs8/src/pem.rs#L42-L86

That implementation has some issues, particularly around its handling of padding (which is both non-constant-time and broken) but that much can be fixed relatively trivially.

Other than that, the actual framing parts can be data-independent.

tarcieri:
Renamed in #222

PR #222, opened 2021-01-22
Rename `b64ct` => `base64ct`

tarcieri:
Per #220, it seems like we need more general constant-time Base64 support:

- PHC "B64" for password hash
- PKCS#8 PEM encoding
- JWK (uses Base64url encoding)

The latter two are (potentially) private key use cases, so they're the ones that would benefit the most from a fully constant-time implementation.

The `b64ct` name comes from a specialization for the first use case. However, now we're using `b64ct` for the latter two cases as well.

This commit renames the crate to `base64ct` with the goal of it covering all of our use cases.


base64ct/Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [package]
-name = "b64ct"
+name = "base64ct"
 version = "0.2.0" # Also update html_root_url in lib.rs when bumping this
```
newpavlov, line 3: Is it worth to do the initial release under v0.2.0 and not v0.1.0?

tarcieri: Oops, I didn't change the version number.

Yeah, I can bump it back down before an initial release.

PR #223, opened 2021-01-22
base64ct: fixups

tarcieri:
Fix version number and description

### DIFF
--- a/pkcs8/Cargo.toml
+++ b/pkcs8/Cargo.toml
@@ -15,7 +15,7 @@ keywords = ["crypto", "key", "private"]
 readme = "README.md"
 
 [dependencies]
-base64ct = { version = "0.2", optional = true, features = ["alloc"], path = "../base64ct" }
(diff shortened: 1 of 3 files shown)


pkcs8/Cargo.toml
```diff
@@ -15,7 +15,7 @@ keywords = ["crypto", "key", "private"]
 readme = "README.md"
 
 [dependencies]
-base64ct = { version = "0.2", optional = true, features = ["alloc"], path = "../base64ct" }
+base64ct = { version = "0", optional = true, features = ["alloc"], path = "../base64ct" }
```
newpavlov, line 18: IIUC `version = "0"` means all pre-1.0 versions, you probably meant 0.0? Also I think it would be better to publish `base64ct v0.1.0`, since 0.0.0 is usually used for empty crates.

tarcieri: Yeah, I'll bump it to `v0.1.0` on release. I'd like to try to upstream the "patterns" being used in `pkcs8` and `elliptic-curve` (for JWT) first and make them constant-time.

PR #225, opened 2021-01-23
base64ct: support both unpadded and padded encodings

tarcieri:
We need support for both:

- "PEM" (as used by PKCS#8) needs padded Base64
- "B64" (as used by the PHC string format) and JWK need unpadded Base64

### DIFF
--- a/base64ct/src/errors.rs
+++ b/base64ct/src/errors.rs
@@ -1,3 +1,5 @@
+//! Error types
+
 use core::fmt;
 
 /// Insufficient output buffer length.
--- a/base64ct/src/lib.rs
+++ b/base64ct/src/lib.rs
@@ -15,7 +15,7 @@
 //! [PHC string format]: https://github.com/P-H-C/phc-string-format/blob/master/phc-sf-spec.md
 //! [RFC 4648, section 4]: https://tools.ietf.org/html/rfc4648#section-4
 
-#![no_std]
+//#![no_std]
 #![cfg_attr(docsrs, feature(doc_cfg))]
 #![doc(
     html_logo_url = "https://raw.githubusercontent.com/RustCrypto/meta/master/logo.svg",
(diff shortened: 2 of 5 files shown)


base64ct/src/base64.rs
```diff
@@ -0,0 +1,229 @@
...
+        return Err(InvalidLengthError);
+    }
+
+    let mut src_offset: usize = 0;
+    let mut dst_offset: usize = 0;
+    let mut src_length: usize = src.len();
+
+    while src_length >= 3 {
+        encode_3bytes(
+            &src[src_offset..(src_offset + 3)],
```
newpavlov, line 25: Why use the `while` loop instead of the approach used in the `b64` module? The latter is more idiomatic and allows compiler to remove bound checks. Same for `decode`.

base64ct/src/base64.rs
```diff
@@ -0,0 +1,229 @@
...
+
+        dst_offset += 4;
+    }
+
+    Ok(dst_offset)
+}
+
+/// Encoded len
+pub fn encoded_len(bytes: &[u8]) -> usize {
+    (((bytes.len() * 4) / 3) + 3) & !3
```
newpavlov, line 54: Remember the overflow bug.

base64ct/src/base64.rs
```diff
@@ -0,0 +1,229 @@
...
+    // SAFETY: `dst` is fully written and contains only valid one-byte UTF-8 chars
+    unsafe { String::from_utf8_unchecked(dst) }
+}
+
+/// Decode
+pub fn decode(src: impl AsRef<[u8]>, dst: &mut [u8]) -> Result<usize, InvalidEncodingError> {
+    let src = src.as_ref();
+
+    // TODO(tarcieri): constant-time whitespace tolerance
+    if !src.is_empty() && char::from(src[src.len() - 1]).is_whitespace() {
```
newpavlov, line 75: Are you sure you want to be so generic instead of simply checking for `b' '`?

base64ct/src/base64.rs
```diff
@@ -0,0 +1,229 @@
...
+
+    ((bytes.len() - pad_count) * 3) / 4
+}
+
+// B64 character set:
+// [A-Z]      [a-z]      [0-9]      +     /
+// 0x41-0x5a, 0x61-0x7a, 0x30-0x39, 0x2b, 0x2f
+
+#[inline(always)]
+pub(crate) fn encode_3bytes(src: &[u8], dst: &mut [u8]) {
```
newpavlov, line 160: Maybe it's worth to move these functions into a separate `utils` module?

base64ct/Cargo.toml
```diff
@@ -14,9 +14,13 @@ categories = ["cryptography", "encoding", "no-std"]
 keywords = ["base64", "phc"]
 readme = "README.md"
 
+[dependencies]
+zeroize = ""
+
```
newpavlov, line 19: Shouldn't zeroization be a concern of a higher level? It's a bit strange to have it in a parsing library.

tarcieri:
@newpavlov the short answer to all of your questions is the code in the `b64ct` crate was derived from a constant-time implementation of `=`-padded Base64, which is what I've used here.

The formulas are slightly different, and as such they do reintroduce previously solved overflow bugs, but that's for "B64" instead of padded Base64.

Also note: work-in-progress 😉

base64ct/src/base64.rs
tarcieri: It should definitely handle newlines as well. They're probably the most common issue here, actually.

All that said, unless there's a separate error variant for whitespace versus other characters, having an eager whitespace check probably provides little value. So we should either add another error variant, or just remove this check entirely.

base64ct/src/base64.rs
tarcieri: I wound up extracting them into a `common` module. There's probably more code that can be DRYed out between them.

base64ct/Cargo.toml
tarcieri: Removed `zeroize`

tarcieri:
I pushed up a slightly different version that only tries to handle the `padded` vs `unpadded` distinction for now.

Still a lot of duplication and unaddressed issues.

base64ct/src/base64.rs
tarcieri: I tried to get this working but had a little bit of trouble. The logic is a bit different from the unpadded case.

base64ct/src/base64.rs
tarcieri: Fixed

tarcieri:
Marking as ready for review.

It's still not great, but the tests are passing and I managed to boil the code down to a single module again.

There's still a lot of repetitiveness, but I ran out of gas trying to DRY it out.

I also didn't try to tackle `decode_in_place` and just renamed it to `decode_in_place_unpadded` for now.

@newpavlov perhaps we can merge this and then you can take a shot at refactoring it?

base64ct/src/base64.rs
tarcieri: I managed to get it working with `chunks` instead of a loop, but I couldn't get it working with `chunks_exact`. Perhaps you could take another look.

tarcieri:
Going to merge this and work on Base64uri support

PR #227, opened 2021-01-25
base64ct: add decode_in_place support for padded Base64

tarcieri:
Previously it was specialized to only unpadded Base64.

Also unifies the padded and unpadded tests, now that they share implementation parity.


base64ct/src/lib.rs
```diff
@@ -202,12 +178,22 @@ pub fn decode<'a>(src: &str, dst: &'a mut [u8], padded: bool) -> Result<&'a [u8]
...
+        let q = (buf.len() - (buf.len() % 4)) / 4;
+
+        if q > 0 {
+            q - 1
+        } else {
+            q
+        }
+    } else {
+        buf.len() / 4
+    };
```
tarcieri, line 196: This seems like a pretty gross way to implement this calculation and I'm not fully confident it's correct. Would love suggestions for an alternative.

base64ct/src/lib.rs
```diff
@@ -287,28 +277,25 @@ pub const fn decoded_len(input: &str, padded: bool) -> usize {
...
     }
 }
 
-#[inline(always)]
-const fn decoded_len_inner_unpadded(n: usize) -> usize {
-    // branchless, overflow-proof computation of `(3*n)/4`
-    let k = n / 4;
-    let l = n - 4 * k;
-    3 * k + (3 * l) / 4
-}
```
tarcieri, line 302: We could potentially keep this extracted as a separate function, but as of now the corresponding padded implementation isn't `const`-friendly.

It would be possible to support a fully const-friendly implementation, but the padded implementation needs to be reworked.

base64ct/src/lib.rs
```diff
@@ -287,28 +277,25 @@ pub const fn decoded_len(input: &str, padded: bool) -> usize {
             i -= 1;
         }
 
+        // TODO(tarcieri): fix potential overflow
         ((bytes.len() - pad_count) * 3) / 4
```
tarcieri, line 281: Oops, just realized I didn't properly address this

base64ct/src/lib.rs
```diff
@@ -202,12 +178,22 @@ pub fn decode<'a>(src: &str, dst: &'a mut [u8], padded: bool) -> Result<&'a [u8]
 }
 
 /// Decode unpadded Base64-encoded string in-place.
-// TODO(tarcieri): support for padded Base64
-pub fn decode_in_place_unpadded(buf: &mut [u8]) -> Result<&[u8], InvalidEncodingError> {
+pub fn decode_in_place(buf: &mut [u8], padded: bool) -> Result<&[u8], InvalidEncodingError> {
```
newpavlov, line 181: It could be worth to mark this function `#[inline]`, to make it easier for compiler propagation of a constant `padded`. Same for other functions which use `padded`.

base64ct/src/lib.rs
newpavlov: But why do you need such complicated computation instead of simply using `(buf.len() / 4).checked_sub(1).unwrap_or(0)`?

tarcieri: Nice! Thank you, that's much more elegant.

I think that will work but (either way) needs a check that `buf.len()` is a multiple of 4.

base64ct/src/lib.rs
tarcieri: Done

base64ct/src/lib.rs
tarcieri: Updated with your suggestion

tarcieri: Ooh: clippy recommends `saturating_sub(1)`. Will keep that in mind for the future.

newpavlov: Heh, I think it's the first time when I encountered a use for a saturating method. :)

BTW I wonder if adding the following code to the begging of decoding functions can be considered as a violation of const-time properties:
```
if padded && buf.len() % 4 != 0 {
   return Err(..);
}
```
Otherwise the length check has to be done as part of the `decode_3bytes_padded` function.

tarcieri: As you mentioned, I think that's fine because it's only variable time in respect to length.

However, there is one part of the code which is data-dependent, which is this loop:

https://github.com/RustCrypto/utils/blob/941e379/base64ct/src/lib.rs#L285-L288

I should add a TODO to fix that (and possibly a GitHub issue)

tarcieri: Added a TODO as part of this PR: https://github.com/RustCrypto/utils/pull/227/files#diff-42d97c4412df722211f6a53e6d276f3d9e7f1f6d33a9e13e9ce94ea6003efd52R278-R279

tarcieri:
@newpavlov look good otherwise? I'd like to work on Base64uri support after this and then follow-up on some of the other TODOs separately from this PR

newpavlov (approved):

PR #228, opened 2021-01-25
base64ct: cleanups

tarcieri:
This commit contains a number of small improvements:

- Removes the remaining data-dependent branches
- Fixes overflow in padded `decoded_len` calculation
- Replaces use of `isize` with `i16` (the specific needed type)
- Extracts methods for constant-time pseudo-branch operations

### DIFF
--- a/base64ct/src/lib.rs
+++ b/base64ct/src/lib.rs
@@ -146,7 +146,7 @@ pub fn decode<'a>(src: &str, dst: &'a mut [u8], padded: bool) -> Result<&'a [u8]
     }
     let src = src.as_bytes();
     let dst = &mut dst[..dlen];
-    let mut err: isize = 0;
+    let mut err: i16 = 0;
 
     if padded {
         for (s, d) in src.chunks(4).zip(dst.chunks_mut(3)) {
@@ -165,7 +165,7 @@ pub fn decode<'a>(src: &str, dst: &'a mut [u8], padded: bool) -> Result<&'a [u8]
         let src_rem = src_chunks.remainder();
         let dst_rem = dst_chunks.into_remainder();
 
-        err |= !(src_rem.is_empty() || src_rem.len() >= 2) as isize;
+        err |= !(src_rem.is_empty() || src_rem.len() >= 2) as i16;
         let mut tmp_out = [0u8; 3];
         let mut tmp_in = [b'A'; 4];
         tmp_in[..src_rem.len()].copy_from_slice(src_rem);
@@ -185,7 +185,7 @@ pub fn decode<'a>(src: &str, dst: &'a mut [u8], padded: bool) -> Result<&'a [u8]
 pub fn decode_in_place(buf: &mut [u8], padded: bool) -> Result<&[u8], InvalidEncodingError> {
     // TODO: eliminate unsafe code when compiler will be smart enough to
     // eliminate bound checks, see: https://github.com/rust-lang/rust/issues/80963
-    let mut err: isize = 0;
+    let mut err: i16 = 0;
     let dlen = decoded_len(&buf, padded);
     let full_chunks = if padded {
         if buf.len() % 4 != 0 {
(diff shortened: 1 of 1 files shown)


base64ct/src/lib.rs
```diff
@@ -364,44 +355,66 @@ fn decode_3bytes(src: &[u8], dst: &mut [u8]) -> isize {
...
+#[inline(always)]
+fn match_byte_ct(input: u8, expected: u8, ret_on_match: i16) -> i16 {
+    match_byte_range_ct(input, expected - 1, expected + 1, ret_on_match)
+}
 
-    // if (ch > 0x2f && ch < 0x3a) ret += ch - 0x30 + 52 + 1; // 5
-    ret += (((47isize - ch) & (ch - 58isize)) >> 8) & (ch + 5isize);
+/// Compute the length of the unpadded portion of a Base64-encoded string
+/// in constant time
+fn unpadded_len_ct(input: &[u8]) -> usize {
```
tarcieri, line 411: It dawns on me that this could be used to replace the current padded decoder with the unpadded decoder, so long as we validated that the padding is well-formed

base64ct/src/lib.rs
```diff
@@ -263,34 +263,25 @@ pub fn decode_vec(input: &str, padded: bool) -> Result<Vec<u8>, Error> {
 }
 
 /// Get the length of the output from decoding the provided Base64-encoded input.
+///
+/// Note that this function does not fully validate the Base64 is well-formed
+/// and may return incorrect results for malformed Base64.
+///
+/// It is designed to be used in conjunction with [`decode`].
```
newpavlov, line 270: Is it worth to keep this function public? Previously it could've been used for creating arrays, but now it's no longer `const`.

base64ct/src/lib.rs
```diff
@@ -364,44 +355,66 @@ fn decode_3bytes(src: &[u8], dst: &mut [u8]) -> isize {
...
+/// in constant time
+fn unpadded_len_ct(input: &[u8]) -> usize {
+    let mut res = 0;
 
-    // if (ch == 0x2b) ret += 62 + 1;
-    ret += (((42isize - ch) & (ch - 44isize)) >> 8) & 63;
+    // TODO(tarcieri): ensure padding bytes are always at the end of a string
+    for &byte in input {
+        res += match_byte_ct(byte, PAD, 1) as usize ^ 1;
+    }
```
newpavlov, line 417: You iterate over the whole string here, which is not ideal. I think it would be better to check if `input` length is larger or equal to 4, and then subtract from the length if last bytes match `PAD`.

newpavlov (commented):
I am not sure if `isize` -> `i16` change is worth the trouble, either way it will be stored in a register without touching the main memory. But since it's already done, I am not opposed to it.

base64ct/src/lib.rs
newpavlov: Maybe introduce a private function which will check the padding and will return `Result<usize, InvalidEncoding>`?

base64ct/src/lib.rs
tarcieri: It's still useful for slicing output buffers

base64ct/src/lib.rs
tarcieri: Sure, that's a better approach

base64ct/src/lib.rs
tarcieri: I was thinking something like that, but instead returning an `err` value which can be OR'd with the other data-dependent values so padding errors are caught in the same branch as any other decoding error

base64ct/src/lib.rs
newpavlov: What do you mean? The `decode` function already returns a sliced buffer.

base64ct/src/lib.rs
tarcieri: Replaced it with subslice matching, which avoids the need to do any length calculations.

[Godbolt output](https://godbolt.org/z/sW6MGT)

tarcieri:
Going to merge this then try to unify padded and unpadded decoding

base64ct/src/lib.rs
tarcieri: Aah yeah, good point. I take back my comment: we could hide this function from the public API.

PR #229, opened 2021-01-26
base64ct: unify padded and unpadded decoding

tarcieri:
Removes padding prior to decoding, allowing the unpadded decoding path to be the "one true way" to handle all decoding.

Validates padding bytes in constant-time, returning an error flag to be OR'd into the other potential encoding errors and branched upon only after fully decoding/validating all other data.

### DIFF
--- a/base64ct/src/errors.rs
+++ b/base64ct/src/errors.rs
@@ -15,29 +15,17 @@ impl fmt::Display for InvalidLengthError {
 #[cfg(feature = "std")]
 impl std::error::Error for InvalidLengthError {}
 
-/// Invalid encoding of provided "B64" string.
-#[derive(Copy, Clone, Debug, Eq, PartialEq)]
-pub struct InvalidEncodingError;
-
-impl fmt::Display for InvalidEncodingError {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
-        f.write_str("invalid B64 encoding")
-    }
-}
-
-#[cfg(feature = "std")]
-impl std::error::Error for InvalidEncodingError {}
-
 /// Generic error, union of [`InvalidLengthError`] and [`InvalidEncodingError`].
 #[derive(Copy, Clone, Debug, Eq, PartialEq)]
-pub enum Error {
+pub enum DecodeError {
     /// Insufficient output buffer length.
     InvalidEncoding,
+
     /// Invalid encoding of provided "B64" string.
     InvalidLength,
 }
 
-impl fmt::Display for Error {
+impl fmt::Display for DecodeError {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
         let s = match self {
             Self::InvalidEncoding => "invalid B64 encoding",
@@ -47,19 +35,12 @@ impl fmt::Display for Error {
     }
 }
 
-impl From<InvalidEncodingError> for Error {
+impl From<InvalidLengthError> for DecodeError {
     #[inline]
-    fn from(_: InvalidEncodingError) -> Error {
-        Error::InvalidEncoding
-    }
-}
-
-impl From<InvalidLengthError> for Error {
-    #[inline]
-    fn from(_: InvalidLengthError) -> Error {
-        Error::InvalidLength
+    fn from(_: InvalidLengthError) -> DecodeError {
+        DecodeError::InvalidLength
     }
 }
 
 #[cfg(feature = "std")]
-impl std::error::Error for Error {}
+impl std::error::Error for DecodeError {}
--- a/base64ct/src/errors.rs
+++ b/base64ct/src/errors.rs
@@ -15,9 +15,22 @@ impl fmt::Display for InvalidLengthError {
 #[cfg(feature = "std")]
 impl std::error::Error for InvalidLengthError {}
 
+/// Invalid encoding of provided "B64" string.
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+pub struct InvalidEncodingError;
+
+impl fmt::Display for InvalidEncodingError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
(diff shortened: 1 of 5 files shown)


base64ct/src/errors.rs
```diff
@@ -15,29 +15,17 @@ impl fmt::Display for InvalidLengthError {
 #[cfg(feature = "std")]
 impl std::error::Error for InvalidLengthError {}
 
-/// Invalid encoding of provided "B64" string.
-#[derive(Copy, Clone, Debug, Eq, PartialEq)]
-pub struct InvalidEncodingError;
```
tarcieri, line 20: There are no longer any cases where this is useful as a separate error anymore, as `decode_in_place` now needs to handle the case where the input is padded but the wrong length.

base64ct/src/lib.rs
```diff
@@ -417,3 +392,36 @@ fn unpadded_len_ct(input: &[u8]) -> usize {
...
+/// errors (i.e. malformed padding bytes) as `i16` to be combined with other
+/// encoding-related errors prior to branching.
+fn validate_padding(input: &[u8]) -> Result<i16, DecodeError> {
+    if input.len() % 4 != 0 {
+        match input.last().cloned() {
+            // TODO(tarcieri): constant-time whitespace check
+            Some(byte) if char::from(byte).is_whitespace() => {
+                return Err(DecodeError::InvalidEncoding)
+            }
+            _ => return Err(DecodeError::InvalidLength),
```
tarcieri, line 408: Not necessarily sure this is a good idea. In fact it seems like the sort of wiggle room that allows things like padding oracle attacks (not saying padding oracle attacks are applicable here)

Perhaps we should just have `EncodeError` and `DecodeError`, or a unit `Error` type?

tarcieri:
Uhoh, looks like a regression in the `pkcs8` crate which isn't covered by the current test suite

base64ct/src/lib.rs
newpavlov: I think it should be enough to simply return `InvalidEncoding`, since incorrect string length is an encoding problem, not buffer length one.

newpavlov (commented):
Maybe it's worth to introduce `*_padded` methods instead of using the `padded` argument? I think it will result in a slightly better code structure and `base64ct::decode_padded(buf, dst)?` is easier to understand in user code than `base64ct::decode(buf, dst, true)?`.

tarcieri:
So it turned out there was the opposite of a regression: the `pkcs8` had some legacy hax to strip padding which the previous implementation allowed and the new implementation rejects (correctly)

> Maybe it's worth to introduce *_padded methods instead of using the padded argument? 

@newpavlov yeah, that sort of thing (which is actually somewhat close to what I used to have) is making more sense as the padded decode at least can be implemented in terms of the padded decode

I'd rather finish implementing Base64uri before refactoring as we don't quite have the picture regarding the complete set of permutations yet

base64ct/src/lib.rs
```diff
@@ -138,65 +141,60 @@ const fn encoded_len_inner(n: usize, padded: bool) -> Option<usize> {
...
-#[inline]
-pub fn decode<'a>(src: &str, dst: &'a mut [u8], padded: bool) -> Result<&'a [u8], Error> {
-    let dlen = decoded_len(src, padded);
+pub fn decode(src: impl AsRef<[u8]>, dst: &mut [u8], padded: bool) -> Result<&[u8], DecodeError> {
+    let mut src = src.as_ref();
+    let mut err = 0;
+
+    if padded {
+        err = validate_padding(src)?;
+        src = &src[..unpadded_len_ct(src)];
```
newpavlov, line 150: You compute `unpadded_len_ct` twice, once here and in `validate_padding`. Maybe it's worth to merge `unpadded_len_ct` and `validate_padding` into a single function?

tarcieri: The only reason why I couldn't unify them into the same function is `decode_vec` needs to call `unpadded_len_ct` in isolation.

If that could be eliminated, I agree it would make more sense to combine them into the same function.

Otherwise I'm not sure how to combine them without duplication.

base64ct/src/lib.rs
tarcieri: That's a potential option, although note that in addition to `InvalidLength` being a potentially confusing error to get for trailing whitespace, it causes a discrepancy between the error returned for trailing whitespace between the padded and unpadded decoding functions, at least given the current implementation.

tarcieri: (ignore my previous reply, I was confused)

Done. This seems ok for now.

base64ct/src/errors.rs
newpavlov: I think wrong padding length falls under `InvalidEncodingError`, not under `InvalidLength`.

base64ct/src/lib.rs
newpavlov: You can remove padding in `decode_vec` before calling `decode`, i.e. you don't have to pass `padded` to decode.

base64ct/src/errors.rs
tarcieri: Okay, but there are still no functions in the public API that return `InvalidEncodingError`, so I don't think it's worth having an extra type for it.

If you really want to keep it around, I think it should be private for now.

base64ct/src/lib.rs
tarcieri: The over-allocation solution is an interesting one

base64ct/src/errors.rs
tarcieri: Added back `InvalidEncodingError` and used it for `decode_in_place` again

base64ct/src/lib.rs
tarcieri: Okay, switched to over-allocation and unified the padding methods into `decode_padding` in a5d0bc4

tarcieri:
Okay, this seems reasonable enough to me now.

@newpavlov feel free to leave additional comments. The implementation of `decode_padding` could probably be improved somewhat, but I'm going to go ahead and merge and try to add Base64url.

PR #231, opened 2021-01-26
base64ct: Base64url encoding support

tarcieri:
Adds support for Base64url: a URL-safe variant of the standard Base64 encoding.

Uses `-` and `_` for encoding Base64 values 62/63, as opposed to the standard `+` and `/`.

### DIFF
--- a/base64ct/src/lib.rs
+++ b/base64ct/src/lib.rs
@@ -12,7 +12,6 @@
 //! Copyright (c) 2014 Steve "Sc00bz" Thomas (steve at tobtu dot com).
 //! Derived code is dual licensed MIT + Apache 2 (with permission from Sc00bz).
 //!
-//! [PHC string format]: https://github.com/P-H-C/phc-string-format/blob/master/phc-sf-spec.md
 //! [RFC 4648, section 4]: https://tools.ietf.org/html/rfc4648#section-4
 
 #![no_std]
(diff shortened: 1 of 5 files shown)


tarcieri:
Quick reference table:

|      | Hex   | Binary       | Alphabet |
|----|------|-----------|----------|
| `+` | 0x2b | 0b101011 | Base64 - 62 |
| `/` | 0x2f | 0b101111 | Base64 - 63 |
| `-` | 0x2d | 0b101101 | Base64url - 62 |
| `_` | 0x5f | 0b1011111 | Base64url - 63 |

The encoder is currently outputting the following for 62/63 instead:

|      | Hex   | Binary       |
|----|------|-----------|
| `)` | 0x29 | 0b101001 |
| `[` | 0x5b | 0b1011011 |

i.e. in either case, the result is off-by-4

base64ct/src/lib.rs
```diff
@@ -269,39 +493,33 @@ pub fn decode_vec(input: &str, padded: bool) -> Result<Vec<u8>, Error> {
...
     let b2 = src[2] as i16;
 
-    dst[0] = encode_6bits(b0 >> 2);
-    dst[1] = encode_6bits(((b0 << 4) | (b1 >> 4)) & 63);
-    dst[2] = encode_6bits(((b1 << 2) | (b2 >> 6)) & 63);
-    dst[3] = encode_6bits(b2 & 63);
+    dst[0] = encode_6bits(b0 >> 2, hi_bytes);
+    dst[1] = encode_6bits(((b0 << 4) | (b1 >> 4)) & 63, hi_bytes);
+    dst[2] = encode_6bits(((b1 << 2) | (b2 >> 6)) & 63, hi_bytes);
+    dst[3] = encode_6bits(b2 & 63, hi_bytes);
```
tarcieri, line 515: My best guess is something needs to change in this code

base64ct/src/lib.rs
```diff
@@ -43,13 +43,228 @@ use alloc::{string::String, vec::Vec};
 /// Padding character
 const PAD: u8 = b'=';
 
+/// Standard Base64 encoding.
+///
+/// ```text
+/// [A-Z]      [a-z]      [0-9]      +     /
+/// 0x41-0x5a, 0x61-0x7a, 0x30-0x39, 0x2b, 0x2f
+/// ```
+pub mod base64 {
```
newpavlov, line 52: Since you introduce modules, I think it could be worth to introduce a `b64` module and move all `unpadded` functions to it while keeping the padded methods in `base64`. It will allow us to remove the `padded`/`unpadded` suffixes, thus improving the naming a bit.

tarcieri: The problem is there are four total combinations this PR supports:

1. Padded standard Base64
2. Unpadded standard Base64
3. Padded URL-safe Base64
4. Unpadded URL-safe Base64

While (2) is "B64", (4) is also an important practical encoding as well: it's used by JWK.

I don't think it would make sense to refer to 4 as "URL-safe B64", as the "B64" name comes from the PHC string format which doesn't define anything for the URL-safe variant.

Another option using modules is to have a structure like:

```rust
mod base64 {
    mod padded {
        ...
    }

    mod unpadded {
        ...
    }
}

mod base64url {
    mod padded {
        ...
    }

    mod unpadded {
        ...
    }
}
```

base64ct/src/lib.rs
```diff
@@ -311,33 +529,27 @@ fn encode_3bytes_padded(src: &[u8], dst: &mut [u8]) {
 }
 
 #[inline(always)]
-fn encode_6bits(src: i16) -> u8 {
+fn encode_6bits(src: i16, hi_bytes: (u8, u8)) -> u8 {
+    let hi_off = 0x1c + (hi_bytes.0 & 4);
```
tarcieri, line 533: I ended up addressing the off-by-4 issue by introducing this tweak.

It works for the Base64/Base64url alphabets but isn't generalizable. I'm also not entirely confident in it, so it'd be good to add some better tests (possibly `proptest` against the `base64` crate)

base64ct/src/lib.rs
newpavlov: How about: `baes64ct::padded`, `baes64ct::unpadded`, `baes64ct::url::padded`, and `baes64ct::url::unpadded`? An alternative to the latter two could be `baes64ct::padded::url` and `baes64ct::unpadded::url`.

tarcieri: Sure, that works

base64ct/src/lib.rs
tarcieri: Would be curious to know if @Sc00bz has a better solution for this problem

base64ct/src/lib.rs
tarcieri: Refactored the code with this module structure

PR #234, opened 2021-01-27
base64ct: minor code improvements

newpavlov:
Improves the `encode` function by improving code reuse and removing an unreachable panic in the padded path. Also marks all inner functions with `#[inline(always)]` to remove call indirection. Most often crates will use only one module, so I think such inlining is warranted.

### DIFF
--- a/base64ct/src/lib.rs
+++ b/base64ct/src/lib.rs
@@ -76,7 +76,7 @@
 #![doc(
     html_logo_url = "https://raw.githubusercontent.com/RustCrypto/meta/master/logo.svg",
     html_favicon_url = "https://raw.githubusercontent.com/RustCrypto/meta/master/logo.svg",
-    html_root_url = "https://docs.rs/base64ct/0.1.0"
+    html_root_url = "https://docs.rs/base64ct/0.1.1"
 )]
 #![warn(missing_docs, rust_2018_idioms)]
 
@@ -335,7 +335,7 @@ pub mod url {
 
 /// Encode the input byte slice as Base64, writing the result into the provided
 /// destination slice, and returning an ASCII-encoded string value.
-#[inline]
+#[inline(always)]
 fn encode<'a>(
     src: &[u8],
     dst: &'a mut [u8],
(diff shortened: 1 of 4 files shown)


base64ct/src/lib.rs
```diff
@@ -353,23 +353,27 @@ fn encode<'a>(
...
-                encode_3bytes_padded(s, d, hi_bytes);
-            }
+        if let Some(dst_rem) = dst_chunks.next() {
+            let mut tmp = [0u8; 3];
+            tmp[..src_rem.len()].copy_from_slice(&src_rem);
+            encode_3bytes(&tmp, dst_rem, hi_bytes);
+
+            let flag = src_rem.len() == 1;
+            let mask = (flag as u8).wrapping_sub(1);
+            dst_rem[2] = (dst_rem[2] & mask) | (PAD & !mask);
```
newpavlov, line 373: Also note that I have replaced the explicit branch with this bit-twiddling. I wonder if there is a better way to generate mask from `bool`.

base64ct/src/lib.rs
```diff
@@ -391,8 +395,9 @@ fn encode<'a>(
 /// If `input` length is greater than `usize::MAX/4`.
 #[cfg(feature = "alloc")]
 #[cfg_attr(docsrs, doc(cfg(feature = "alloc")))]
+#[inline(always)]
 fn encode_string(input: &[u8], padded: bool, hi_bytes: (u8, u8)) -> String {
-    let elen = encoded_len(input, padded);
+    let elen = encoded_len_inner(input.len(), padded).expect("input is too big");
```
newpavlov, line 400: Thanks to the forced inlining, the second `expect` used on `encode` result will be properly removed.

base64ct/src/lib.rs
```diff
@@ -353,23 +353,27 @@ fn encode<'a>(
...
+    let src_rem = src_chunks.remainder();
+
     if padded {
-        for (s, d) in src.chunks(3).zip(dst.chunks_mut(4)) {
-            if s.len() == 3 {
-                encode_3bytes(s, d, hi_bytes);
-            } else {
-                encode_3bytes_padded(s, d, hi_bytes);
-            }
+        if let Some(dst_rem) = dst_chunks.next() {
```
tarcieri, line 366: Oh neat, I guess that was all I was missing.

tarcieri (approved):

PR #236, opened 2021-01-31
base64ct: refactor encoder functions

tarcieri:
To make it easier to support the "crypt" encoding (which is slightly different from the others as the alphabet is shifted), this change refactors encoders to support an arbitrary `encode_6bits` function per-encoding.


tarcieri:
Given this requires a function-per-encoding, I'm beginning to wonder if it might be helpful to have an `Encoding` trait

PR #237, opened 2021-01-31
base64ct: bcrypt encoding

tarcieri:
Adds the Base64 variant used by bcrypt


tarcieri:
Note: I was originally trying to add the `crypt(3)` encoding for SHA-crypt, only to discover that `crypt(3)` and bcrypt use slightly different encodings(!)

Issue #263, opened 2021-02-17
pkcs8: decryption/encryption support for EncryptedPrivateKeyInfo


tarcieri:
#262 added an initial `pkcs8::EncryptedPrivateKeyInfo` type with basic parsing/serialization support. However, it doesn't actually support decrypting/encrypting `PrivateKeyInfo` yet.

Ideally we should only support algorithms which are known to be secure. The most commonly supported ones are based on 56-bit DES, however those provide no effective security as 56-bit DES has far too small a keyspace to be secure against brute force attacks. However, there is support for modern algorithms like AES and old-but-still-secureish algorithms like 3DES in newer revisions of PKCS#5:

- [PKCS#5 v1.5] supports `PBE-SHA1-3DES`.
- [PKCS#5 v2] adds support for AES encryption with iterated PRFs such as `hmacWithSHA256` (a.k.a. PBES2)

It would probably make sense to wait for the [`cipher` crate v0.3 release](https://github.com/RustCrypto/traits/issues/392) before attempting to implement decryption/encryption support in `pkcs8`.

[PKCS#5 v1.5]: https://tools.ietf.org/html/rfc2898
[PKCS#5 v2]: https://tools.ietf.org/html/rfc8018

tarcieri:
It seems like the first step towards getting this to work is implementing at least a subset of PKCS#5 (password-based encryption), which I've done in #272.

tarcieri:
This PR adds support for PKCS5 decryption: https://github.com/RustCrypto/utils/pull/293

tarcieri:
Encryption is support is released in v0.5.3.

PR #272, opened 2021-02-17
pkcs5: initial crate

tarcieri:
Adds an initial implementation of Public-Key Cryptography Standards (PKCS) 5: Password-Based Cryptography Specification Version 2.1 as described in [RFC 8018](https://tools.ietf.org/html/rfc8018).

This implementation includes basic support for parsing `AlgorithmIdentifier`s representing supported PKCS#5 algorithms into a structured representation.

The larger goal is being able to decrypt/encrypt PKCS#8's `EncryptedPrivateKeyInfo` (see #263).

This PR doesn't actually include any encryption support yet: instead it implements the basic crate structure with the goal of adding encryption support in follow-up PRs.

It also includes some changes to the `spki` crate to simplify parsing which should probably be separated out into their own PR.


pkcs5/src/pbes2.rs
```diff
@@ -0,0 +1,223 @@
...
+/// Pseudo-random function used by PBKDF2.
+// TODO(tarcieri): support default `algid-hmacWithSHA1`?
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+#[non_exhaustive]
+pub enum Pbkdf2Prf {
+    /// HMAC with SHA-256
+    HmacWithSha256,
+
+    /// Other PRFs
+    Other(ObjectIdentifier),
```
tarcieri, line 165: Several PBES2 enums include these catch-all `Other` variants, as this PR only implements a single ciphersuite (PBKDF2-SHA256 w\ AES-256-CBC).

It would be good exhaustively implement all of the algorithms described in RFC 8018, however even then the RFC says "Applications may employ other schemes as well" so it seems these should probably remain.

newpavlov: What is a hypothetical use-case for those `Other` variants instead of simply returning an "unknown algorithm" error immediately? It looks a bit strange to have `#[non_exhaustive]` with `Other` variant.

tarcieri: Per the RFC, it'd be an application-specific use case. Note that this one in particular is fairly deep within the structure.

Simply returning an error in these cases is an alternative, although that said, the `der` crate doesn't presently have a good way to propagate an "unknown OID" error (which is probably something that should be added, and can be done in a non-breaking way).

tarcieri: Went ahead and removed the `Other` variants and now return `der::ErrorKind::OidInvalid` (added in #273) which wraps the invalid OID value and returns it back to the caller.

The remaining `enum`s in the `pbes2` module are all marked `#[non_exhaustive]` so we can incrementally add additional algorithm support (although especially for PBES2, I'd like to also add test vectors for each algorithm we add)

PR #284, opened 2021-02-19
const-oid: fix root arc calculation

tarcieri:
It was using bitwise OR instead of addition.

This commit also significantly expands the tests, including adding test cases for a root arc of 2, assertions for the root arcs everywhere (as they are represented differently internally), and additionally moves tests into the `tests/` directory so we don't need conditional `std` linkage for the tests.

### DIFF
--- a/pkcs5/src/pbes2.rs
+++ b/pkcs5/src/pbes2.rs
@@ -22,7 +22,7 @@ pub const HMAC_WITH_SHA256_OID: ObjectIdentifier =
 /// 256-bit Advanced Encryption Standard (AES) algorithm with Cipher-Block
 /// Chaining (CBC) mode of operation.
 pub const AES_256_CBC_OID: ObjectIdentifier =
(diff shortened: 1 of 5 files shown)


pkcs5/src/pbes2.rs
```diff
@@ -22,7 +22,7 @@ pub const HMAC_WITH_SHA256_OID: ObjectIdentifier =
 /// 256-bit Advanced Encryption Standard (AES) algorithm with Cipher-Block
 /// Chaining (CBC) mode of operation.
 pub const AES_256_CBC_OID: ObjectIdentifier =
-    ObjectIdentifier::new(&[2, 0, 840, 1, 101, 3, 4, 1, 42]);
+    ObjectIdentifier::new(&[2, 16, 840, 1, 101, 3, 4, 1, 42]);
```
tarcieri, line 25: I've confirmed this is the correct OID for this algorithm:
http://www.oid-info.com/cgi-bin/display?oid=2.16.840.1.101.3.4.1.42&action=display

PR #293, opened 2021-02-21
pkcs5: encryption

### LINKED ISSUE #263: pkcs8: decryption/encryption support for EncryptedPrivateKeyInfo

tarcieri:
Implements PKCS#5 encryption support, presently targeting only support for PBES2 with PBKDF2-SHA-256 and AES-CBC (with 128 or 256-bit key size).

Note that these algorithm choices are presently the best options supported by PKCS#5 v2.1.

Closes #263


tarcieri:
Note: this initial draft is encryption only, but it manages to pass OpenSSL-generated test vectors! 🎉 

The `aes` and `block-modes` crates are the latest as sourced from git (and in that regard, it looks like this feature will require an MSRV bump, which is presently causing the tests to fail).

I'd like to wait until we release `cipher` v0.3 along with accompanying releases of `aes` (and whatever will happen to `block-modes`, i.e. rolled into `cipher`?)

All that said, I can probably finish out adding things like encryption support as well as integrating this functionality with the `pkcs8` crate in this PR as well in the intervening time.

tarcieri:
Due to the MSRV bump, I backpedaled on using the latest prerelease `aes` and `block-modes` crates in ff44ba6.

The upgrade is simple and I'd like to cut another breaking `der` release anyway which will also necessitate a minor version bump.

I'm pretty close to finishing this out and can cut an initial release with support, then we can circle back on the MSRV bump and crate upgrades.

PR #312, opened 2021-02-24
const-oid: const-friendly string parser

tarcieri:
Adds a new `ObjectIdentifier::parse` function which is declared `const fn` and capable of parsing OIDs from dot-delimited strings in const contexts.

This is much more readable/ideal than the current approach used by `ObjectIdentifier::new`, which accepts a slice of `Arc` (i.e. `u32`) type instead.

Ideally we can phase out the current `::new` in favor of one which parses strings, however that'd be a breaking change, so for now the implementation is side-by-side with the current `::new`.

### DIFF
--- a/const-oid/src/parser.rs
+++ b/const-oid/src/parser.rs
@@ -1,0 +1,162 @@
+use crate::{Arc, ObjectIdentifier, MAX_ARCS};
+
+/// Constant panicking assertion.
+// TODO(tarcieri): use const panic when stable.
+// See: https://github.com/rust-lang/rust/issues/51999
+macro_rules! const_assert {
+    ($bool:expr, $msg:expr) => {
+        [$msg][!$bool as usize]
+    };
+}
+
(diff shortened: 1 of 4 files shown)


const-oid/src/parser.rs
```diff
@@ -0,0 +1,162 @@
...
+        b'6' => 6,
+        b'7' => 7,
+        b'8' => 8,
+        b'9' => 9,
+        other => {
+            const_assert!(matches!(other, b'0'..=b'9'), "invalid ASCII digit");
+            0 // Unreachable due to above `const_assert`
+        }
+    }
+}
```
tarcieri, line 133: Oof, wrote this locally using `saturating_sub` only to discover that it was made `const fn` in Rust 1.47 and the current MSRV is 1.46. So we need these unfortunate hacks for now.

`const fn` is a bit bleeding edge but rapidly getting better 😅

PR #315, opened 2021-02-26
block-buffer: add block_mode_processing methods

newpavlov:
Also adds the `get_full_block` method to `LazyBlockBuffer`.

### DIFF
--- a/block-buffer/src/buffer.rs
+++ b/block-buffer/src/buffer.rs
@@ -1,11 +1,11 @@
 #[cfg(feature = "block-padding")]
 use block_padding::Padding;
 
-use core::slice;
 use crate::{
     utils::{to_blocks, to_blocks_mut},
-    Block, DigestBuffer, ParBlock,
+    Block, DigestBuffer, InvalidLength, ParBlock,
 };
+use core::slice;
 use generic_array::{typenum::U1, ArrayLength};
 
 /// Buffer for block processing of data.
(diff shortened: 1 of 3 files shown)


newpavlov:
I plan to double-check the new code a bit later, but otherwise this PR should be good to merge.

block-buffer/src/buffer.rs
```diff
@@ -93,6 +94,65 @@ impl<BlockSize: ArrayLength<u8>> BlockBuffer<BlockSize> {
...
+    pub fn block_mode_processing<'a>(
+        &mut self,
+        mut data: &[u8],
+        buf: &'a mut [u8],
+        mut process: impl FnMut(&mut [Block<BlockSize>]),
+    ) -> Result<&'a [u8], InvalidLength> {
+        let pos = self.get_pos();
+        let rem = self.remaining();
+        let mut blocks_processed = 0;
+        let (_, mut buf_blocks, _) = to_blocks_mut::<BlockSize, U1>(buf);
```
tarcieri, line 109: Perhaps this instead?

```suggestion
        let mut buf_blocks = to_blocks_mut::<BlockSize, U1>(buf).1;
```

block-buffer/src/lazy.rs
```diff
@@ -9,6 +13,65 @@ pub struct LazyBlockBuffer<BlockSize: ArrayLength<u8>> {
...
+    pub fn block_mode_processing<'a>(
+        &mut self,
+        mut data: &[u8],
+        buf: &'a mut [u8],
+        mut process: impl FnMut(&mut [Block<BlockSize>]),
+    ) -> Result<&'a [u8], InvalidLength> {
+        let pos = self.get_pos();
+        let rem = self.remaining();
+        let mut blocks_processed = 0;
+        let (_, mut buf_blocks, _) = to_blocks_mut::<BlockSize, U1>(buf);
```
tarcieri, line 28: [Same thing here](https://github.com/RustCrypto/utils/pull/315/files#r586485755)?

tarcieri (approved):

PR #325, opened 2021-03-04
[WIP] der: replace typenum with const generics for BigUInt

tarcieri:
BigUInt was previously using `typenum` to define the size of an integer. This is necessary because it maps to a signed ASN.1 `INTEGER` and needs to add/remove a leading 0 when the first octet is `>0x7f`.

This is perhaps the one thing in RustCrypto right now where I feel that `min_const_generics` stabilization might actually help: it's a trivial usage, doesn't involve `generic-array` or associated constants of a trait definition, it's just using the type system to be generic around an integer value.

This eliminates `typenum` as a dependency and with it feature gating `big-uint`. Presently requires `beta` until 1.51 ships on March 25th.

### DIFF
--- a/der/src/asn1/big_uint.rs
+++ b/der/src/asn1/big_uint.rs
@@ -3,8 +3,7 @@
 use crate::{
     Any, ByteSlice, Encodable, Encoder, Error, ErrorKind, Header, Length, Result, Tag, Tagged,
 };
-use core::{convert::TryFrom, marker::PhantomData};
-use typenum::Unsigned;
+use core::convert::TryFrom;
 
 /// "Big" unsigned ASN.1 `INTEGER` type.
 ///
@@ -19,16 +18,12 @@ use typenum::Unsigned;
 ///
 /// Currently supported sizes are 1 - 512 bytes.
 #[derive(Copy, Clone, Debug, Eq, PartialEq)]
-#[cfg_attr(docsrs, doc(cfg(feature = "big-uint")))]
-pub struct BigUInt<'a, N: BigUIntSize> {
+pub struct BigUInt<'a, const N: usize> {
     /// Inner value
     inner: ByteSlice<'a>,
-
-    /// Integer size in bytes
-    size: PhantomData<N>,
 }
 
-impl<'a, N: BigUIntSize> BigUInt<'a, N> {
+impl<'a, const N: usize> BigUInt<'a, N> {
     /// Create a new [`BigUInt`] from a byte slice.
     ///
     /// Slice may be less than or equal to `N` bytes.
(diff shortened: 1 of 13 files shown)


der/src/asn1/big_uint.rs
```diff
@@ -145,64 +137,19 @@ impl<'a, N: BigUIntSize> Encodable for BigUInt<'a, N> {
...
-    U370, U371, U372, U373, U374, U375, U376, U377, U378, U379, U380, U381, U382, U383, U384, U385,
-    U386, U387, U388, U389, U390, U391, U392, U393, U394, U395, U396, U397, U398, U399, U400, U401,
-    U402, U403, U404, U405, U406, U407, U408, U409, U410, U411, U412, U413, U414, U415, U416, U417,
-    U418, U419, U420, U421, U422, U423, U424, U425, U426, U427, U428, U429, U430, U431, U432, U433,
-    U434, U435, U436, U437, U438, U439, U440, U441, U442, U443, U444, U445, U446, U447, U448, U449,
-    U450, U451, U452, U453, U454, U455, U456, U457, U458, U459, U460, U461, U462, U463, U464, U465,
-    U466, U467, U468, U469, U470, U471, U472, U473, U474, U475, U476, U477, U478, U479, U480, U481,
-    U482, U483, U484, U485, U486, U487, U488, U489, U490, U491, U492, U493, U494, U495, U496, U497,
-    U498, U499, U500, U501, U502, U503, U504, U505, U506, U507, U508, U509, U510, U511, U512
-);
```
tarcieri, line 195: This was really a roundabout way of disallowing `U0`. In hindsight I'm not sure it matters.

newpavlov: `typenum` has the [`NonZero`](https://docs.rs/typenum/1.12.0/typenum/marker_traits/trait.NonZero.html) marker trait for such use cases. :)

tarcieri: Oh neat!

newpavlov:
I am not sure it's worth to migrate only some crates while most of other crates have to depend on `typenum`. Plus it could be worth for `der` depend on the hypothetical bigint crate, which with high probability will use `typenum`. Nevertheless I will leave the final decision to you.

tarcieri:
I would like to double check that it's possible to use a `typenum` constant as a const generic parameter, as that's the main use case for this right now (via the `ecdsa` crate).

If I can't get that to work, there's no point, but if it does work out I think it's worth a shot.

I'm also working on trying to implement ASN.1 SET OF and think that could greatly benefit from const generics as well.

newpavlov:
>I would like to double check that it's possible to use a typenum constant as a const generic parameter

Hm, since `USIZE` is an associated constant of the [`Unsigned`](https://docs.rs/typenum/1.12.0/typenum/marker_traits/trait.Unsigned.html) trait, you may encounter the same issue which does not allow us to use arrays dependent on a trait associated constant.

tarcieri:
Alas, appears you're right. Guess we'll have to wait until const generics evolve a bit more.

Screen Shot 2021-03-05 at 7 46 23 AM

PR #334, opened 2021-03-17
base64ct: MSRV 1.47 and TODO fixes it enables

tarcieri:
All of the TODOs except for one are because `const fn` support for various methods isn't available until 1.47, and the previous MSRV was 1.46 instead.

Therefore, this commit bumps MSRV from 1.46 to 1.47 (which will entail a minor version bump too)

### DIFF
--- a/base64ct/src/encoding.rs
+++ b/base64ct/src/encoding.rs
@@ -90,10 +90,9 @@ impl<T: Variant> Encoding for T {
         }
     }
 
-    /// Decode a standard Base64 string without padding in-place.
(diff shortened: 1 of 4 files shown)


base64ct/src/encoding.rs
```diff
@@ -296,16 +292,14 @@ fn is_pad_ct(input: u8) -> i16 {
 
 #[inline(always)]
 const fn encoded_len_inner(n: usize, padded: bool) -> Option<usize> {
-    // TODO: replace with `checked_mul` and `map` on stabilization
```
tarcieri, line 299: 1.47+ lets us use `checked_mul`, but unfortunately `map` is not `const fn` in any release to date

PR #335, opened 2021-03-17
base64ct v1.0.0

tarcieri:
### Changed
- Bump MSRV to 1.47+ ([#334])

### Fixed
- MSRV-dependent TODOs in implementation ([#334])

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -33,7 +33,7 @@ dependencies = [
 
 [[package]]
 name = "base64ct"
-version = "0.2.1"
+version = "1.0.0"
 
 [[package]]
 name = "blobby"
--- a/base64ct/CHANGELOG.md
+++ b/base64ct/CHANGELOG.md
@@ -4,6 +4,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## 1.0.0 (2021-03-17)
+### Changed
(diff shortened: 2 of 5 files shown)


newpavlov:
Hm, generally I would be more conservative about releasing v1.0. It's not only about API stability, but also about having MSRV-dependent selection of dependency versions. But I guess it should be fine for this crate in particular.

tarcieri:
I'd consider this crate effectively "done".

Not sure what you mean by "MSRV-dependent selection of dependency versions"... minor 1.x versions? FWIW, I don't anticipate needing to bump MSRV on this crate for the foreseeable future.

newpavlov:
>I'd consider this crate effectively "done".

It's just an extra precaution, even if a crate is "done" we may find some tweaks later, which would be nice to add. Even in a relatively small crate like `subtle` they had to release v2.0 just 3 months after v1.0 release. This is why I think it's reasonable to have several months of "cool down" period before releasing v1.0. Thanks to the semver-trick such delayed release can be quite painless for migration.

>Not sure what you mean by "MSRV-dependent selection of dependency versions"... minor 1.x versions?

I meant that MSRV provided by a crate will be used by `cargo` during dependency version resolution as [described](https://github.com/rust-lang/rfcs/blob/master/text/2495-min-rust-version.md#influencing-version-resolution) in the RFC, i.e. if v1.0 has MSRV 1.47 and 1.1 has MSRV 1.60, `cargo` will select v1.0 if compiler has version smaller than 1.60. This would allow us to update crate MSRVs without releasing v2.0. Unfortunately it does not look like this feature will be implemented anytime soon.

PR #365, opened 2021-04-03
when processing hex input, ignore comments behind #

mcr:
This allows hex literals to contains comments behind #...
This works well with testing things that deal with CBOR content, such as processed by cbor.me.
I have a short test case, but it can not go into the same crate.
Shall I create another crate?
Perhaps this should be a different macro?

### DIFF
--- a/hex-literal/src/lib.rs
+++ b/hex-literal/src/lib.rs
@@ -83,16 +83,28 @@ impl TokenTreeIter {
     }
 
     fn next_hex_val(&mut self) -> Option<u8> {
+        let mut incomment = false;
         loop {
             let v = match self.buf.get(self.pos) {
                 Some(&v) => v,
                 None => return None,
             };
             self.pos += 1;
+            if incomment {
+                match v {
+                    b'\n' => { incomment = false; }
+                    _ => { /* nothing */ }
+                };
+                continue;
+            }
             let n = match v {
                 b'0'..=b'9' => v - 48,
                 b'A'..=b'F' => v - 55,
                 b'a'..=b'f' => v - 87,
+                b'#' => {
+                    incomment = true;
+                    continue;
+                }
                 b' ' | b'\r' | b'\n' | b'\t' => continue,
                 _ => panic!("encountered invalid character"),
             };


tarcieri:
I'd personally like some way of adding comments to hex literals to have "literate" hex literals. Not sure about this syntax specifically, but to illustrate what I'd like:

```rust
hex!("
    AA BB // header
          04 // length
                DE AD BE EF
");
```

mcr:
> I'd personally like some way of adding comments to hex literals to have "literate" hex literals. Not sure about this syntax specifically, but to illustrate what I'd like:

That's why I suspect that we might want multiple different formats as to what's acceptable.
I just ran into a problem though, one can't have \a in the comment, as the \ gets parsed earlier, and may be an invalid escape.
While CBOR output looks like:

```
         D9 0104                        # tag(260)
            50                          # bytes(16)
               FE80000000000000609C62FFFED8ABBA # "\xFE\x80\x00\x00\x00\x00\x00\x00`\x9Cb\xFF\xFE\xD8\xAB\xBA"
```
which I'd like to be able paste literally in. (see cbor.me, or cbor2pretty.rb)

tarcieri:
Yeah didn't mean to imply `#` was a bad alternative over `//`. Your example looks good too.

mcr:
> Yeah didn't mean to imply `#` was a bad alternative over `//`. Your example looks good too.

I'd want to have a few hex_FOO! variants, which might imply this should be done differently.
Also, it seems that we can't have " in the comment area, so maybe regarding it all as a string is also wrong.
(This is my first attempt to write a macro)

mcr:
Should I create a new crate for putting tests into?  My code has more problems than my trivial initial testing showed.

tarcieri:
> Also, it seems that we can't have " in the comment area, so maybe regarding it all as a string is also wrong.

Comments can be embedded in the string literal potentially. That's what I was illustrating in my examples.

It's not the only way to go though... removing the string literal quoting would allow comments to be syntax highlighted more appropriately depending on the editor.

hex-literal/src/lib.rs
```diff
@@ -83,16 +83,28 @@ impl TokenTreeIter {
...
         loop {
             let v = match self.buf.get(self.pos) {
                 Some(&v) => v,
                 None => return None,
             };
             self.pos += 1;
+            if incomment {
+                match v {
+                    b'\n' => { incomment = false; }
+                    _ => { /* nothing */ }
```
newpavlov, line 96: I think this `match` can be replaced with a single line:
```rust
incomment = v == b'\n';
```

It (and your variant) may be somewhat incorrect for characters which contain `\n` in their UTF-8 encoding, but I guess it's better to simply forbid non-ASCII comments for now.

newpavlov (commented):
Looks good! Can you please also add tests with comments and update documentation?

I think it would be nice to have a compatible syntax for comments with the rest of Rust (i.e. `//` instead of `#`), but I understand that `#` is a bit simpler to parse.

>I have a short test case, but it can not go into the same crate.

Have you tried only inline module tests? The existing doctest works without issues, I also think that `tests/` folder should work as well.

tarcieri:
FWIW I'm also okay with `#` as the comment syntax. Don't want to bikeshed too much.

mcr:
#512 does not accept # comments.
I suppose that I can add this.
The reason why I want # comments, is because that's what cbor.me aka cbor2diag.rb produces.

PR #367, opened 2021-04-15
der: fix length overflow on add

tarcieri:
In PR #349, the internal representation of a Length was changed from `u16` to `u32` to make it easier to support documents larger than 64kB in the future.

However, this added a `Length::max` invariant which wasn't getting properly checked when doing addition (it was only checking that the primitive `u32` integer addition operation didn't overflow).

This change adds a test which captured the overflow behavior, and also adds a `TryFrom<u32> for Length` impl which ensures the length is in-range, and changes the rest of the length-related code to always use this impl to instantiate `Length` types.

### DIFF
--- a/der/src/length.rs
+++ b/der/src/length.rs
@@ -27,6 +27,7 @@ impl Length {
     }
 
     /// Get the maximum length supported by this crate
+    // TODO(tarcieri): support lengths greater than 65535
     pub const fn max() -> Self {
         Self(65535)
     }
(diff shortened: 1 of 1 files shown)


tarcieri:
Note: as far as I can tell this is not a security-critical bug. Integer overflows which would cause wrapping behavior are still properly caught using `checked_add`.

The actual impact of this bug is it would allow generation of a `Length` which would fail to properly encode. That encoding failure would generate an `ErrorKind::Overflow` at the time data was encoded: https://github.com/RustCrypto/utils/blob/23997c812dd7ed94c9b336394d7c6c569743ea31/der/src/length.rs#L191-L206

PR #381, opened 2021-04-26
Rename `cpuid-bool` to `cpufeatures`

tarcieri:
If we plan on expanding this crate to include ARM targets, it seems like a more general name would make sense.


tarcieri:
Going to move forward with this

PR #393, opened 2021-05-05
cpufeatures: aarch64 support (Linux and macOS/M4)

### LINKED ISSUE #378: cpuid-bool: ARM support
It would be nice to be able to do ARM feature detection.

A notable use case would be detecting the [ARMv8 Cryptography Extensions](https://github.com/RustCrypto/asm-hashes/issues/5#issuecomment-812890980).

tarcieri:
Adds preliminary support for runtime feature detection on `aarch64` targets, presently restricted to the following set of target features which are present on both `aarch64-unknown-linux-gnu` and `aarch64-apple-darwin` targets:

- `aes`: AES support
- `sha2`: SHA1 and SHA256 support
- `sha3`: SHA512 and SHA3 support

Closes #378

### DIFF
--- a/cpufeatures/src/aarch64.rs
+++ b/cpufeatures/src/aarch64.rs
@@ -1,0 +1,112 @@
+//! ARM64 CPU feature detection support.
+//!
+//! Unfortunately ARM instructions to detect CPU features cannot be called from
+//! unprivileged userspace code, so this implementation relies on OS-specific
+//! APIs for feature detection.
+
+#[macro_export]
+#[doc(hidden)]
+macro_rules! __unless_target_features {
+    ($($tf:tt),+ => $body:expr ) => {
+        {
+            #[cfg(not(all($(target_feature=$tf,)*)))]
+            $body
+
+            #[cfg(all($(target_feature=$tf,)*))]
+            true
+        }
+    };
+}
+
+#[cfg(target_os = "linux")]
+#[macro_export]
+#[doc(hidden)]
+macro_rules! __detect_target_features {
+    ($($tf:tt),+) => {{
+        let hwcaps = unsafe { libc::getauxval(libc::AT_HWCAP) };
+        $($crate::check!(hwcaps, $tf) & )+ true
+    }};
+}
+
+#[cfg(target_os = "macos")]
+#[macro_export]
+#[doc(hidden)]
+macro_rules! __detect_target_features {
+    ($($tf:tt),+) => {{
+        $($crate::check!($tf) & )+ true
+    }};
+}
+
+/// Linux `expand_check_macro`
+#[cfg(target_os = "linux")]
+macro_rules! __expand_check_macro {
+    ($(($name:tt, $hwcap:expr)),* $(,)?) => {
+        #[macro_export]
+        #[doc(hidden)]
+        macro_rules! check {
+            $(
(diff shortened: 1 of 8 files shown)


cpufeatures/src/aarch64.rs
```diff
@@ -0,0 +1,91 @@
...
+                ($hwcaps:expr, $name) => { (($hwcaps & libc::$name) != 0) };
+            )*
+        }
+    };
+}
+
+expand_check_macro! {
+    ("aes",  HWCAP_AES),  // Enable AES support.
+    ("sha2", HWCAP_SHA2), // Enable SHA1 and SHA256 support.
+    ("sha3", HWCAP_SHA3), // Enable SHA512 and SHA3 support.
```
tarcieri, line 90: Note: the comments here are taken directly from the output of:

```
$ rustc --target=aarch64-unknown-linux-gnu --print target-features
```

I'm a bit confused because the target features aren't isomorphic to either the Linux ARM64 ELF hwcaps or the macOS sysctl flags.

Per the Linux ARM64 ELF hwcaps docs:

https://www.kernel.org/doc/html/latest/arm64/elf_hwcaps.html

> HWCAP_SHA1: Functionality implied by ID_AA64ISAR0_EL1.SHA1 == 0b0001.
> HWCAP_SHA2: Functionality implied by ID_AA64ISAR0_EL1.SHA2 == 0b0001.
> HWCAP_SHA3: Functionality implied by ID_AA64ISAR0_EL1.SHA3 == 0b0001.
> HWCAP_SHA512: Functionality implied by ID_AA64ISAR0_EL1.SHA2 == 0b0010.

That would suggest that the SHA-1 and SHA-3 hwcaps are independent of the others, and that SHA-2 support implies SHA-512 support.

macOS/M4 likewise has separate `sysctl(3)` flags for SHA-512 and SHA-3:

```
hw.optional.armv8_2_sha512: 1
hw.optional.armv8_2_sha3: 1
```

tarcieri:
Note: the current implementation is copy/paste between x86(_64) and `aarch64`.

I'd like to get it working at all on the latter, and then can circle back on refactoring/DRYing out the implementation.

tarcieri:
Marking this as ready for review.

I was able to extract the CPU-specific bits from the `new!` macro and factor them into respective modules, so the core macro that sets up the `AtomicU8`s and caches the detection results is shared between targets.

Linux and MacOS targets now work on `aarch64`. I configured CI to test a Linux target using `cross`, and tested a MacOS target locally on a M1 Mini.

cpufeatures/src/aarch64.rs
```diff
@@ -0,0 +1,112 @@
+//! ARM64 CPU feature detection support.
+//!
+//! Unfortunately ARM instructions to detect CPU features cannot be called from
+//! unprivileged userspace code, so this implementation relies on OS-specific
+//! APIs for feature detection.
+
+#[macro_export]
+#[doc(hidden)]
+macro_rules! __unless_target_features {
+    ($($tf:tt),+ => $body:expr ) => {
```
tarcieri, line 10: Played around with a few different syntaxes for this macro and this is the best one I could come up with.

Open to suggestions for alternatives.

tarcieri:
Going to go ahead and merge this, but I'll hold off on publishing the changes until next week to give @newpavlov some time to review.

I'll move my concerns about the mappings of ARM CPU features Rust target features to a separate issue.

PR #403, opened 2021-05-13
cpufeatures: fix usage on Musl environments; add NEON support

tarcieri:
The `libc` crate doesn't presently define Linux's `HWCAP_*` in Musl-based environments:

https://github.com/rust-lang/libc/issues/2171

This PR (temporarily) copies the presently supported `HWCAP_*` constants directly into this crate.

Additionally, it adds support for the `neon` target feature.


tarcieri:
It seems adding runtime detection for NEON to `aarch64` is unnecessary as it can be assumed to be present.

It would be nice for 32-bit ARM, however.

PR #413, opened 2021-05-17
crypto-bigint

tarcieri:
This commit adds an experimental prototype const generics-based big integer library targeting cryptographic use cases.

It's an extraction of some of the code from the `elliptic-curve` crates, modified to use const generics.

The immediate intended use case is being able to use it for elliptic curve constants, such as a curve's `Order`:

https://docs.rs/elliptic-curve/0.9.11/elliptic_curve/trait.Order.html

As such, it's been written with liberal usage of `const fn`. This allows safe usage in const scenarios which isn't possible with `generic-array`.

All that said, in general I think we could benefit from a library like this for several applications, including things like `rsa` and `dsa`.

### DIFF
--- a/.github/workflows/crypto-bigint.yml
+++ b/.github/workflows/crypto-bigint.yml
@@ -1,0 +1,55 @@
+name: crypto-bigint
+
+on:
+  pull_request:
+    paths:
+      - "crypto-bigint/**"
+      - "Cargo.*"
+  push:
+    branches: master
+
+defaults:
+  run:
+    working-directory: crypto-bigint
+
+env:
+  CARGO_INCREMENTAL: 0
+  RUSTFLAGS: "-Dwarnings"
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        rust:
+          - 1.51.0 # MSRV
+          - stable
+        target:
+          - thumbv7em-none-eabi
+          - wasm32-unknown-unknown
+    steps:
+      - uses: actions/checkout@v1
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: ${{ matrix.rust }}
+          target: ${{ matrix.target }}
+          override: true
+      - run: cargo build --target ${{ matrix.target }} --release
+
+  test:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        rust:
+          - 1.51.0 # MSRV
+          - stable
+    steps:
+      - uses: actions/checkout@v1
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: ${{ matrix.rust }}
+          override: true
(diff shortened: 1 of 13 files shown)


fjarri:
Interesting, I started to work on a [similar library](https://github.com/fjarri/dark-integers) recently, to see if the compiler is strong enough to handle all the required type magic.

nickray:
Still quite experimental (playing with what I'd like to have and what's possible with current const generics), I'm working on a similar thing for a more targeted use case: https://github.com/ycrypto/rsa-cortex-m4 (currently on hold, main missing part is prime generation). Ideally, I'd like a "mathematical" interface like PARI/GP or Sage, but reasonably efficient in embedded contexts. Algorithmically, I'm trying to implement ideas from https://www.shiftleft.org/papers/rsacrt/ and elsewhere. 

The main difference is probably that I use **two** const usizes, D "digits" and E "extra digits", in `Unsigned`, which is `#[repr(C)] struct { d: [u8; D], e: [u8; E] }`. The idea there being that the two primes of an RSA key have E = 0, and multiply to a number with E = D (corresponding to `Short` and `Long` integers). The problem being that [u8; 2 * const] is not currently a thing... I do also have an (unsigned integer) `Number` trait that might be helpful generally after some scrutiny. Goal is to stay within stable Rust in any case.

From the `core::ops` side, I think all three of
- non-truncating (leading to a `Product<D, E, N>` type)
- wrapping
- modulo prime or odd or power of two
should be exposed in a nice way, without getting lost in far too much generality in the common cases.

The other question I'm pursuing is how to model "nice" properties of numbers (e.g. being odd, having the top bit set for a given limb count, being prime) and then take advantage of them, as there's a multitude of specialized algorithms ("Arazi's lemma", partial reduction while calculating with modular integers etc.) that would be nice to dispatch depending on a number's properties. Currently doing this with nested newtypes, abusing `AsRef`.

And then there's the question of what to implement in constant time :) Something like "iterating over all limbs in basic arithmetic, even if they're zero", is probably useful to do in any case; other things might be major performance losses and better implemented in non-constant time with appropriate masking. Probably don't want too many feature knobs to turn.

tarcieri:
> Algorithmically, I'm trying to implement ideas from https://www.shiftleft.org/papers/rsacrt/ and elsewhere.

That was definitely a part of the plan here, along with potentially sourcing certain algorithms from BearSSL.

I also agree about having (potentially layered) newtypes. We're already doing that sort of thing in the `elliptic-curve` crate, but I think what we're doing can be even further decomposed into types provided by this crate.

> And then there's the question of what to implement in constant time

I'm trying to use constant-time operations as a baseline, with `subtle` as a hard dependency. However, from there it's possible to go from `Choice` to `bool` or `CtOption` with an `Into` conversion.

I think in many cases it will be nice to have both options, but perhaps the constant-time version can be the easy-to-use one, and the variable time the more verbose one (e.g. `mul_vartime`)

PR #425, opened 2021-05-24
pkcs8: expose Attributes

tarcieri:
Exposes raw `Attributes` for the `OneAsymmetricKey` struct.

No attempt is made to parse them as a `SET OF Attribute`. Instead it exposes an inner `Any` value for the context-specific tag 0.

The main use case for this is likely to be an optional certificate field, possibly used in the context of PKCS#12, however we presently have no examples with attributes which makes it difficult to model them any more specifically than a wrapper for an `Any` value.

This should be enough that if someone actually needs access to the attributes that they're exposed and available.

cc @ShadowJonathan

### DIFF
--- a/pkcs8/src/attributes.rs
+++ b/pkcs8/src/attributes.rs
@@ -1,31 +1,70 @@
+//! `OneAsymmetricKey` attributes.
+
 use core::convert::TryFrom;
 
-use der::{Any, Encodable, Encoder, Error, Length, Result, Tag, Tagged};
+use der::{Any, Encodable, Encoder, Error, Header, Length, Result, Tag, Tagged};
 
-// TODO: make proper attributes struct
-#[derive(Debug)]
-pub(crate) struct Attributes;
+/// Attributes as defined in [RFC 5958 Section 2]:
+///
+/// > attributes is OPTIONAL.  It contains information corresponding to
+/// > the public key (e.g., certificates).  The attributes field uses the
+/// > class `ATTRIBUTE` which is restricted by the
+/// > `OneAsymmetricKeyAttributes` information object set.
+/// > `OneAsymmetricKeyAttributes` is an open ended set in this document.
+/// > Others documents can constrain these values.  Attributes from
+/// > RFC2985 MAY be supported.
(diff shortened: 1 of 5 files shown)


pkcs8/src/attributes.rs
```diff
@@ -1,31 +1,70 @@
...
+    pub fn from_raw_attrs(attrs: Any<'a>) -> Self {
+        Self { inner: attrs }
+    }
 
-// TODO: make proper attributes struct
-#[derive(Debug)]
-pub(crate) struct Attributes;
+    /// Borrow the inner attributes value as an [`Any`].
+    // TODO(tarcieri): decode this as a `SetOf` and allow iteration?
+    pub fn as_any(&self) -> &Any<'a> {
```
ShadowJonathan, line 42: maybe better to have `inner()` instead?

pkcs8/src/attributes.rs
```diff
@@ -1,31 +1,70 @@
...
+///
+/// ```text
+/// Attributes ::= SET OF Attribute { { OneAsymmetricKeyAttributes } }
+/// ```
+///
+/// [RFC 5958 Section 2]: https://datatracker.ietf.org/doc/html/rfc5958
+#[derive(Copy, Clone, Debug, Eq, PartialEq, PartialOrd, Ord)]
+pub struct Attributes<'a> {
+    /// Inner ASN.1 value (i.e. `SET OF Attribute`).
+    // TODO(tarcieri): decode this as a `SetOf`?
```
ShadowJonathan, line 27: it's very unlikely that attributes would exist in the wild, i'd say it's okay to leave this as undefined as it is here as a semi-scream-test, so to see if someone has trouble decoding a particular document because it has attributes, so it's okay for now.

tarcieri: I think attributes may actually be used to store the certificate in the context of PKCS#12 (which is an extension of PKCS#8v2), but I'm uncertain.

ShadowJonathan: nw nw

PR #426, opened 2021-05-24
pkcs8: add TODO for proper extension handling; fix dups

tarcieri:
The ellipsis `...` in the ASN.1 schema was previously interpreted as allowing duplicate fields for `attributes` and `public_key`, however it seems that it should correctly be interpreted as an extension marker which allows extension fields to exist after `attributes` and `public_key`.

It seems that, properly handled, extension fields should be ignored, which we're not doing yet, so this is marked as a TODO for now. The `der` crate should likely provide a better mechanism for skipping over unknown extension fields.

cc @ShadowJonathan

### DIFF
--- a/pkcs8/src/one_asymmetric_key.rs
+++ b/pkcs8/src/one_asymmetric_key.rs
@@ -87,24 +87,17 @@ impl<'a> TryFrom<der::Any<'a>> for OneAsymmetricKey<'a> {
             let algorithm = decoder.decode()?;
             let private_key = decoder.octet_string()?.into();
 
-            let mut attributes = None;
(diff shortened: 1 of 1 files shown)


ShadowJonathan:
Re, `der` support: maybe a decoder option that tries to decode `Any` variants until one successfully converts/decodes into the type wanted? `.skip_until<T>() -> Option<T>`

tarcieri:
I was thinking more like adding a proper `ContextSpecific` type, and then doing a match on the tag, ignoring fields with unknown context-specific tags.

But I think for now this is fine and we can circle back on that later.

PR #435, opened 2021-05-26
cpufeatures: add iOS support (static ARM64 capabilities only)

tarcieri:
All Apple ARM64 hardware has the same baseline set of statically known capabilities which can be assumed on all iOS (and macOS) platforms.

This commit adds support for those statically known capabilities on iOS.

Unfortunately it does not appear to be possible to access the `sysctl(3)` namespace on iOS in order to determine the availability of other CPU features which aren't part of this baseline set the same way we can on macOS, so a static capability set is the best we can do.

See this issue for more information: #378

### DIFF
--- a/cpufeatures/src/aarch64.rs
+++ b/cpufeatures/src/aarch64.rs
@@ -135,8 +135,31 @@ pub unsafe fn sysctlbyname(name: &[u8]) -> bool {
     value != 0
 }
 
+// iOS `check!` macro.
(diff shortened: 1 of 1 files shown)


jrose-signal:
Can you tag a release of this one too, please? :-)

tarcieri:
@jrose-signal released in `cpufeatures` v0.1.5: #502

PR #456, opened 2021-06-02
cpufeatures: map `aarch64` HWCAPs to target features; add `crypto`

### LINKED ISSUE #395: cpufeatures: incongruities between ARM64 hwcaps and Rust SHA* target features
Following up from #393:

The output of `rustc --target=aarch64-unknown-linux-gnu --print target-features` shows the following cryptography-related target features on ARM64 (which is the same for `aarch64-apple-darwin`)

```
    sha2                               - Enable SHA1 and SHA256 support.
    sha3                               - Enable SHA512 and SHA3 support.
```

This suggests that `sha2` implies SHA-1 support, and that `sha3` implies SHA(2)-512 support (but curiously, that the aforementioned `sha2` does not).

This does not consistent with the Linux hwcaps or the ARM registers they map to (or for that matter, the algorithm families the respective algorithms belong to):

https://www.kernel.org/doc/html/latest/arm64/elf_hwcaps.html

```
HWCAP_SHA1: Functionality implied by ID_AA64ISAR0_EL1.SHA1 == 0b0001.
HWCAP_SHA2: Functionality implied by ID_AA64ISAR0_EL1.SHA2 == 0b0001.
HWCAP_SHA3: Functionality implied by ID_AA64ISAR0_EL1.SHA3 == 0b0001.
HWCAP_SHA512: Functionality implied by ID_AA64ISAR0_EL1.SHA2 == 0b0010.
```

On macOS targets, support for `sha1` and `sha2` is implicit, however support for SHA-512 and SHA-3 intrinsics can be queried through `sysctl(3)`.

Question:

How should `cpufeatures` handle this mapping? For example, should checking for support for the `sha3` target feature using `cpufeatures` test for support for both SHA(2)-512 and SHA-3, and return false unless both are available?

Sidebar: do the Rust target feature mappings actually make sense here? Should `sha3` perhaps be decoupled from SHA(2)-512?

tarcieri:
Closes #395

The HWCAPs provided by Linux are finer-grained than LLVM target features. This crate is trying to detect the latter, so we need to group together these finer grained features detected via HWCAPs.

This commit attempts to group together finer grained HWCAPs into coarser grained target features, ensuring all of the relevant HWCAPs for a given target feature are available.

Additionally, this adds support for the `crypto` target feature, which implies `aes` (and with it PMULL), along with `sha2` and `neon` (although the latter is always present on `aarch64`)

### DIFF
--- a/cpufeatures/Cargo.toml
+++ b/cpufeatures/Cargo.toml
@@ -15,7 +15,7 @@ edition = "2018"
 readme = "README.md"
 
 [target.aarch64-apple-darwin.dependencies]
-libc = "0.2"
+libc = "0.2.95"
 
 [target.'cfg(all(target_arch = "aarch64", target_os = "linux"))'.dependencies]
-libc = "0.2"
+libc = "0.2.95"
--- a/cpufeatures/src/aarch64.rs
+++ b/cpufeatures/src/aarch64.rs
@@ -66,21 +66,27 @@ macro_rules! __expand_check_macro {
 // Linux `expand_check_macro`
 #[cfg(target_os = "linux")]
 __expand_check_macro! {
-    ("aes",  HWCAP_AES),  // Enable AES support.
-    ("sha2", HWCAP_SHA2), // Enable SHA1 and SHA256 support.
-    ("sha3", HWCAP_SHA3), // Enable SHA512 and SHA3 support.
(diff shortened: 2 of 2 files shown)


jrose-signal:
Looks like this is missing "crypto" for iOS?

PR #512, opened 2021-07-01
Line and block comments.

### LINKED ISSUE #511: Feature request, allow comments in hex-literal::hex macro
I would like to be able to use comments like so:

```rust
   hex!(
       "
        // header
        0001 006c 2112a44238656d797950694b78506e6e
        // username
        0006 0025 63636431623031363037000000
        // etc
        c057 0004 00010032 // or here
     ");
```

As `/` is not currently a valid character, it could be added without causing any breaking changes.

`//` would just mean ignore all characters until the next line.

If you are open to this idea, I am happy to make a PR.

alan-signal:
Closes #511

Closes #365 ('#' prefixed comments)

I kept the comment removal as a separate module so that the tests can be scoped to just removing commented parts of a string. There only needed to be a few documentation tests to show the macro was working as intended.

I decided to do `/* */` style comments too because it was not much more work, but I can remove those states from the machine easily enough if you really do not want them.

### DIFF
--- a/hex-literal/src/comments.rs
+++ b/hex-literal/src/comments.rs
@@ -1,0 +1,272 @@
+//! Provides an Iterator<Item=u8> decorator that uses a finite state machine to exclude comments
+//! from a string in linear time and constant space.
+
+use std::iter::Peekable;
+
+pub(crate) trait Exclude: Sized + Iterator<Item = u8> {
+    fn exclude_comments(self) -> ExcludingComments<Self>;
+}
+
+impl<T: Iterator<Item = u8>> Exclude for T {
+    fn exclude_comments(self) -> ExcludingComments<T> {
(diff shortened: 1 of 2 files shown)


tarcieri (approved):
Looks good to me!

@newpavlov any thoughts?

newpavlov (approved):
Thank you, looks great!

mcr:
I guess that I will send a pull requests against this to accept # as a line comment.

newpavlov:
I am not sure that we want to support both `#` and `//` for line comments. As I wrote in #365 I think it's worth to be consistent with Rust in this regard.

mcr:
Hi, so the point of importing hex with comments is so that one can import output from other things to use as unit test data.
If I have to edit things, even if I do that with a sed script, then writing unit test cases gets harder.

PR #553, opened 2021-07-26
pkcs1: support for customizing PEM `LineEnding`

tarcieri:
Adds `*pem_with_le` methods that support customizing the `LineEnding` used when encoding PEM.

### DIFF
--- a/pkcs1/src/document/private_key.rs
+++ b/pkcs1/src/document/private_key.rs
@@ -11,7 +11,10 @@ use zeroize::{Zeroize, Zeroizing};
 
 #[cfg(feature = "pem")]
 use {
-    crate::{pem, private_key::PEM_TYPE_LABEL},
+    crate::{
+        pem::{self, LineEnding},
+        private_key::PEM_TYPE_LABEL,
+    },
     alloc::string::String,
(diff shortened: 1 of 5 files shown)


pkcs1/src/document/private_key.rs
```diff
@@ -86,8 +89,8 @@ impl ToRsaPrivateKey for RsaPrivateKeyDocument {
 
     #[cfg(feature = "pem")]
     #[cfg_attr(docsrs, doc(cfg(feature = "pem")))]
-    fn to_pkcs1_pem(&self) -> Result<Zeroizing<String>> {
-        let pem_doc = pem::encode_string(PEM_TYPE_LABEL, Default::default(), self.as_der())?;
+    fn to_pkcs1_pem_with_le(&self, line_ending: LineEnding) -> Result<Zeroizing<String>> {
```
dignifiedquire, line 92: I know it is too late, but this method name really makes me think of little endian, not line ending unfortuantely

tarcieri: Can file an issue about better naming. Here you go:

https://github.com/RustCrypto/utils/issues/558

PR #510, opened 2021-06-27
experimental: add modulo operations to UInt

dignifiedquire:
Implements

- `add_mod`
- `sub_mod`
- `neg_mod`
- `mul_mod`

for `UInt<1..=12>` manually

PR #561, opened 2021-07-29
crypto-bigint: add modulo operations to `UInt`

tarcieri:
Continuing on @dignifiedquire's work from #510, this PR attempts to simplify the implementation.

~~Aspirationally, the goal is to add equivalence tests with a known-good implementation.~~ Proptests against `num-bigint` added.


crypto-bigint/src/uint/add_mod.rs
```diff
@@ -1,123 +1,97 @@
...
-        impl UInt<$size> {
-            /// Computes `a + b mod p` in constant time.
-            ///
-            /// Assumes `a` and `b` are `< p`.
-            pub const fn add_mod(&self, rhs: &Self, p: &Self) -> Self {
-                base::$base_name(&self, &rhs, &p)
-            }
+use crate::{Limb, UInt};
+
+impl<const LIMBS: usize> UInt<LIMBS> {
```
dignifiedquire, line 5: Given that we do not have specialization yet, this now prevents me from adding optimizations for specific limb sizes, which are critical for optimal performance

tarcieri: What if we have a generic `const fn` version that always works, and traits which are specific to limb sizes?

That's how all of the existing macros in this crate work anyway... they write trait impls.

dignifiedquire: fair, that should work as well

tarcieri:
Added some trait wrappers in 2b37f48. A couple notes:

- used macros to impl them for `1..=12` limbs, but you can potentially omit some of those and add different impls
- they have the same name as the `const fn` which is a bit annoying/confusing, but has mostly worked out in the elliptic curve crates
- I added `debug_assert`s for `< p`

Ideally we can eventually replace that with `const impl` and specialization in some hypothetical future.

tarcieri:
@dignifiedquire I tried to write a proptest for `add_mod` in 54a3c47, but it's failing:

https://github.com/RustCrypto/utils/pull/561/checks?check_run_id=3196567313

However the code looks correct to me, so I'm not sure if there's something wrong with the test itself or the code.

PR #563, opened 2021-07-29
crypto-bigint: add initial proptests for equivalence with num-bigint

tarcieri:
Adds an initial set of property-based tests to check equivalence between `crypto_bigint::U256` and `num_bigint::BigUint`.

I mostly want these for #561, but I figured I could push up an initial integration separately.

### DIFF
--- a/crypto-bigint/src/uint/mul.rs
+++ b/crypto-bigint/src/uint/mul.rs
@@ -44,7 +44,7 @@ impl<const LIMBS: usize> UInt<LIMBS> {
 
     /// Perform wrapping multiplication, discarding overflow.
     pub const fn wrapping_mul(&self, rhs: &Self) -> Self {
-        self.mul_wide(rhs).0
+        self.mul_wide(rhs).1
     }
 
     /// Perform checked multiplication, returning a [`CtOption`] which `is_some`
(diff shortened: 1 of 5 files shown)


crypto-bigint/src/uint/mul.rs
```diff
@@ -44,7 +44,7 @@ impl<const LIMBS: usize> UInt<LIMBS> {
 
     /// Perform wrapping multiplication, discarding overflow.
     pub const fn wrapping_mul(&self, rhs: &Self) -> Self {
-        self.mul_wide(rhs).0
+        self.mul_wide(rhs).1
```
tarcieri, line 47: As it were, these tests uncovered a bug

tarcieri:
Ugh, looks like it's breaking due to `resolver = "1"`-style feature unification

dignifiedquire (approved):
nice

tarcieri:
Will need to factor it into a pseudo-crate for now to avoid breaking `no_std` CI

crypto-bigint/src/uint/mul.rs
newpavlov: We probably should cut a release with this fix and yank the versions with this bug. Even though I've seen this comment, I recently hit this bug and it took me some time to uncover this bug.

PR #572, opened 2021-08-08
fix: set constructed bit in SET

### LINKED ISSUE #571: SET OF asn.1 type should always have the constructed bit set
According to section 5.15 of the layman's guide (https://luca.ntop.org/Teaching/Appunti/asn1.html) and the let's encrypt asn.1 documentation (https://letsencrypt.org/docs/a-warm-welcome-to-asn1-and-der/#tag), bit 6 of the SET OF tag (the constructed bit) should always be set to 1. This is similar to the SEQUENCE tag. It currently is outputted as 0x17 rather than 0x31.

ajpauwels:
fixes #571


tarcieri:
Thank you!

PR #579, opened 2021-08-16
crypto-bigint: implement Hash

newpavlov:
We probably should implement other `std` traits as well.

@tarcieri 
Are there other traits in your opinion in addition to `Eq` with which we should be careful?


tarcieri:
> Are there other traits in your opinion in addition to Eq with which we should be careful?

`Eq`/`PartialEq` and `Ord`/`PartialOrd` are both implemented in terms of the `subtle` traits.

That said, I'm a little concerned with `Hash` too. Internally it should use SipHash by default which should be ok, but it does provide a sort of sharp edge where it could be used with other hashes which could potentially do data-dependent branching.

I guess it's probably okay.

PR #583, opened 2021-08-20
Use from_bytes_with_nul for string check

kornelski:
`CStr` is doing exactly what the assert is meant to do, and gives the right char type by default.


tarcieri:
Thanks!

PR #612, opened 2021-09-03
crypto-bigint: `Integer` trait

tarcieri:
Adds a trait encompassing the interesting bounds of `UInt`.

Since traits like `elliptic_curve::Curve` can't yet be generic over `const LIMBS`, this trait provides a convenient way of wrapping up `UInt` behavior such that it's still usable without spelling out all of these bounds explicitly.

### DIFF
--- a/crypto-bigint/src/traits.rs
+++ b/crypto-bigint/src/traits.rs
@@ -1,6 +1,81 @@
 //! Traits provided by this crate
 
 use crate::Limb;
+use core::fmt::Debug;
+use subtle::{
+    Choice, ConditionallySelectable, ConstantTimeEq, ConstantTimeGreater, ConstantTimeLess,
+};
+
+#[cfg(feature = "rand")]
+use rand_core::{CryptoRng, RngCore};
+
+/// Integer type.
+pub trait Integer:
+    'static
+    + AsRef<[Limb]>
+    + Copy
+    + ConditionallySelectable
+    + ConstantTimeEq
+    + ConstantTimeGreater
+    + ConstantTimeLess
+    + Debug
+    + Default
+    + Encoding
+    + Eq
+    + From<u64>
+    + Ord
+    + Send
+    + Sized
+    + Sync
+{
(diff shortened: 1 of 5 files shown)


crypto-bigint/src/traits.rs
```diff
@@ -1,6 +1,81 @@
...
+use core::fmt::Debug;
+use subtle::{
+    Choice, ConditionallySelectable, ConstantTimeEq, ConstantTimeGreater, ConstantTimeLess,
+};
+
+#[cfg(feature = "rand")]
+use rand_core::{CryptoRng, RngCore};
+
+/// Integer type.
+pub trait Integer:
```
dignifiedquire, line 13: why no requirements for Add, Mul etc?

tarcieri: Those traits are presently only impl'd for `Wrapping<UInt<_>>` in order to be explicit about the wrapping vs checked behavior.

tarcieri: FWIW, in `elliptic-curve` I end up bounding on the modular arithmetic traits: https://github.com/RustCrypto/traits/pull/732/files#diff-58bf09549ae77b47fcc1f18c3335fd0f6a4eb49a1e3cee51332fe40dcfd66daaR113-R120

```rust
type UInt: bigint::AddMod<Output = Self::UInt>
        + bigint::Integer
        + bigint::NegMod<Output = Self::UInt>
        + bigint::Random
        + bigint::RandomMod
        + bigint::SubMod<Output = Self::UInt>
        + zeroize::Zeroize;
```

They could potentially be included in the `Integer` bounds.

tarcieri: I guess I didn't add bound on `*Mod` initially because they are only impl'd for a subset of the supported integer values

dignifiedquire: Yeah makes sens to wait for adding those bound, until we have a more complete set

PR #619, opened 2021-09-08
crypto-bigint: fix `add_mod` overflow handling

tarcieri:
For an as yet undetermined reason I can't implement scalar arithmetic for the `k256`/`p256` crates using the current `add_mod` implementation (or rather, the test vectors fail).

It appears to have something to do with overflow handling, and how the current implementation unconditionally subtracts the modulus regardless of whether or not the `adc` operation overflowed.

However, while the implementation in this commit allows the `k256`/`p256` tests to pass, it fails all of the current proptests.

cc @dignifiedquire @fjarri @str4d @tuxxy

### DIFF
--- a/crypto-bigint/proptests/tests/equivalence.rs
+++ b/crypto-bigint/proptests/tests/equivalence.rs
@@ -1,6 +1,6 @@
 //! Equivalence tests between `num-bigint` and `crypto-bigint`
 
-use crypto_bigint::{Encoding, U256, Limb};
+use crypto_bigint::{Encoding, U256};
 use num_bigint::BigUint;
 use proptest::prelude::*;
 use std::mem;
@@ -39,7 +39,7 @@ proptest! {
     fn roundtrip(a in uint()) {
         assert_eq!(a, to_uint(to_biguint(&a)));
     }
-    
+
     #[test]
     fn wrapping_add(a in uint(), b in uint()) {
         let a_bi = to_biguint(&a);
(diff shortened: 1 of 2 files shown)


fjarri:
I'm not sure I understand the checks being done in `test_add()`. You have 
```rust
let x = a.wrapping_add(&b);
if x < *p {
    assert_eq!(c, x, "incorrect result");
}
```
But if `a + b` does wrap around the maximum integer size, won't the `assert_eq` fail? I actually added the values on which the test fails manually (in Python), and `c` turned out to be the correct result, yet that assertion fired.

dignifiedquire:
@fjarri you are right, there is a check missing that `a` and `b` are both at most half of the max size.

tarcieri:
So it seems we need to fix both those tests and the proptests?

tarcieri:
I went ahead and merged the test from #623 into this branch, and removed the other tests as it seems they're incorrectly written.

That's not a great state of affairs, but it should at least fix the implementation and have at least one test until we can add better tests.

I can make a tracking issue for adding back tests, but in the meantime I'd like to merge this as it does seem to fix a bug.

tarcieri:
I opened #626 to track adding back some additional tests

tarcieri:
I can confirm that with this fix `k256` and `p256` can now use `add_mod`: https://github.com/RustCrypto/elliptic-curves/pull/426

dignifiedquire:
I have some ideas about fixing the tests locally, will try to pr some things tomorrow.

PR #624, opened 2021-09-12
more tests for shl

mikelodder7:
Also fixes issue when shift % BIT_SIZE == 0.

### DIFF
--- a/crypto-bigint/src/uint/shl.rs
+++ b/crypto-bigint/src/uint/shl.rs
@@ -1,9 +1,10 @@
 //! [`UInt`] bitwise left shift operations.
 
 use super::UInt;
-use crate::limb::BIT_SIZE;
+use crate::limb::{Inner, BIT_SIZE};
+use crate::uint::bits::{ct_select, is_nonzero};
 use crate::Limb;
-use core::ops::Shl;
+use core::ops::{Shl, ShlAssign};
 
 impl<const LIMBS: usize> UInt<LIMBS> {
     /// Computes `self << shift`.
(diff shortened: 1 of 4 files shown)


crypto-bigint/src/uint/shl.rs
```diff
@@ -21,17 +22,20 @@ impl<const LIMBS: usize> UInt<LIMBS> {
         }
 
         let shift_num = n / BIT_SIZE;
-        let shift_rem = n % BIT_SIZE;
+        let rem = n % BIT_SIZE;
+        let nz = is_nonzero(rem as Inner);
+        let lshift_rem = rem as Inner;
+        let rshift_rem = ct_select(0, (BIT_SIZE - rem) as Inner, nz);
```
tarcieri, line 28: It seems like you could use `subtle` for this:

```suggestion
        let lshift_rem = rem as Inner;
        let rshift_rem = Inner::conditional_select(0, (BIT_SIZE - n) as Inner), !rem.ct_eq(&0));
```

mikelodder7: Subtle is not a const fn

mikelodder7: In order to have const fn for shift you have to use these. I would prefer subtle than reinventing the wheel

mikelodder7: Once traits support const fn in stable then we can remove these

tarcieri: Aha, makes sense

PR #625, opened 2021-09-12
Add DES and Triple DES support to PKCS5

jklong:
* Adds PBES2 DES/DES3 tests to pkcs5 and pkcs8
* Implements DES-CBC and DES-EDE3-CBC PBES2 decryption

This is needed to decrypt PKCS8 files generated with legacy versions of the OpenSSL CLI circa 1.0.2 which has seen widespread use due to FIPS 140-2  validation. The default settings for the pkcs8 CLI in this version used DES-EDE3-CBC for compatibility reasons before finally moving to AES256-CBC in v1.1.0.

All DES functionality is gated behind a set of _des_ or _pbes2-des_ features.


tarcieri:
Overall this looks OK.

I'm a bit wary of including 56-bit (non-triple) DES without a fairly large warning that it's insecure. 3DES seems ok.

Perhaps they could be under separate crate features? (e.g. `3des` and `des-insecure`)

jklong:
Yeah that's fair. I'll redo the features and add some large warnings around it being for legacy support only and not recommended for any new data

jklong:
That CI failure looks to  be an issue with the runner.

```

  stable-x86_64-unknown-linux-gnu updated - rustc 1.55.0 (c8dfcfe04 2021-09-06) (from rustc 1.54.0 (a178d0322 2021-07-26))

info: checking for self-updates
error: could not download file from 'https://static.rust-lang.org/rustup/release-stable.toml' to '/tmp/rustup-update7dhPyw/release-stable.toml': failed to make network request: error sending request for url (https://static.rust-lang.org/rustup/release-stable.toml): connection error: Connection reset by peer (os error 104): connection error: Connection reset by peer (os error 104): Connection reset by peer (os error 104)
Error: The process '/usr/share/rust/.cargo/bin/rustup' failed with exit code 1
```

pkcs5/src/pbes2.rs
```diff
@@ -2,6 +2,8 @@
 //!
 //! [RFC 8018 Section 6.2]: https://tools.ietf.org/html/rfc8018#section-6.2
 
+#![cfg_attr(docsrs, feature(doc_cfg))]
```
tarcieri, line 5: This is only needed in `lib.rs`

tarcieri (approved):
This looks good aside from an unneeded `doc_cfg`

PR #630, opened 2021-09-13
Add Error case for PEM with headers fields

dwhjames:
PEM from RFC 1421 defines header fields
https://datatracker.ietf.org/doc/html/rfc1421.html#section-9
this change detects header fields as an initial line that contains the
colon char.

PEM from RFC 7468 disallows
https://datatracker.ietf.org/doc/html/rfc7468#section-2

> Unlike legacy PEM encoding [RFC1421], OpenPGP ASCII armor, and the
> OpenSSH key file format, textual encoding does *not* define or permit
> headers to be encoded alongside the data.

---

The motivation here is to be able to distinguish PEM documents that may valid under RFC 1421 rules, but are definitely not under RFC 7468 rules.


dwhjames:
@tarcieri this is related to
- https://github.com/str4d/rage/pull/260
- https://github.com/str4d/rage/pull/261

To be able to recognize legacy OpenSSH encrypted RSA private keys, I thought it would be helpful to be able to distinguish between a corrupted PEM document and one with headers.

tarcieri:
The general idea of returning a different error when headers are encountered in this PR is ok.

However, in its current form this crate is carefully written to avoid branching directly on the content of the "encapsulated text" in the event the input document is well-formed.

This PR, as written, appears to branch on every single character of the first line of the encapsulated text, regardless of if it parses successfully as Base64.

I think it would be fine to re-examine the first line if it fails to parse as Base64, and return a different error if it appears to be a header.

pem-rfc7468/src/decoder.rs
```diff
@@ -177,21 +180,35 @@ impl<'a> Iterator for Lines<'a> {
...
-        } else if !self.bytes.is_empty() {
-            let line = self.bytes;
-            self.bytes = &[];
-            Some(Ok(line))
+        if self.is_start
+            && self
+                .bytes
+                .iter()
+                .take_while(|&&b| !grammar::is_vwsp(b))
+                .any(|&b| b == grammar::CHAR_COLON)
```
tarcieri, line 188: This is where branching appears to be occurring on the encapsulated text

dwhjames:
> I think it would be fine to re-examine the first line if it fails to parse as Base64, and return a different error if it appears to be a header. But always branching on every character of the encapsulated test, even for the first line (some keys such as Ed25519 ones are only one line), defeats the goals of this crate.

@tarcieri understood; I took another crack at this

tarcieri:
Sorry to do this but you happened to catch me in the middle of migrating this crate to a new repository:

https://github.com/RustCrypto/formats/tree/master/pem-rfc7468

Would you mind reopening this PR there?

dwhjames:
NP; https://github.com/RustCrypto/formats/pull/13

PR #632, opened 2021-09-13
Implement Divide and Modulo crypto-bigint

mikelodder7:
Implements Constant Time Divide and Remainder (Modulo) operations for BigInt

### DIFF
--- a/crypto-bigint/proptests/Cargo.toml
+++ b/crypto-bigint/proptests/Cargo.toml
@@ -13,4 +13,5 @@ members = ["."]
 [dependencies]
 crypto-bigint = { path = ".." }
 num-bigint = "0.4"
+num-traits = "0.2"
 proptest = "1"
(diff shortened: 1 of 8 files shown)


mikelodder7:
Hmm might need some help with prop tests since num-bigint panics on zero but mine doesn’t

tarcieri:
Looks good, thank you!

I have a few nits but also would really like to get `crypto-bigint` moved to a new home at:

https://github.com/rustcrypto/crypto-bigint

In the interest of getting this moved over I'm going to go ahead and merge this, and open some follow-ups there after the crate has been moved.

mikelodder7:
Excellent. Feel free to let me know what to do to make it better

tarcieri:
I was mostly concerned about nitpicking method names and things that maybe shouldn't be in the public API (yet)

mikelodder7:
Right. Makes sense

PR #633, opened 2021-09-14
expand algorithm coverage for PBES2-PBKDF2

dwhjames:
- recognize all sha1 and sha2 hmac psuedo-random functions
- add aes-192-cbc
- test coverage for all combinations


tarcieri:
@dwhjames mind rebasing? I just merged #625

pkcs5/Cargo.toml
```diff
@@ -31,7 +32,7 @@ hex-literal = "0.3"
 
 [features]
 alloc = []
-pbes2 = ["aes", "block-modes", "hmac", "pbkdf2", "scrypt", "sha2"]
+pbes2 = ["aes", "block-modes", "hmac", "pbkdf2", "scrypt", "sha-1", "sha2"]
```
tarcieri, line 35: Ideally I think it'd be possible to use the `pbes2` feature with a ciphersuite free of any questionable cryptographic primitives. My personal preference would be to make SHA-1 separately optional, possibly something like this:

```toml
sha1 = ["pbes2", "sha-1"]
```

dwhjames: ok; this is at least hmacWithSha1 as opposed to plain Sha1

your suggestion is to do something along the lines of what was just done in PR #625, i.e., sprinkle the appropriate
```#[cfg(feature = …``` around?

dwhjames: Ah, so hmacWithSha1 is the default algorithm

https://github.com/RustCrypto/utils/blob/0fe702320c124abcce9425bf83f40ee0c159f781/pkcs5/src/pbes2/kdf.rs#L232-L244

One of the fixes in this PR related to
https://github.com/RustCrypto/utils/blob/0fe702320c124abcce9425bf83f40ee0c159f781/pkcs5/src/pbes2/encryption.rs#L139-L142

What would you propose to do here?
https://github.com/RustCrypto/utils/blob/f1de48a1f96bd7b38451cb5146ef041a63096100/pkcs5/src/pbes2/encryption.rs#L130-L135
return CryptoError as before?

dwhjames: revised in https://github.com/RustCrypto/utils/pull/633/commits/5bb661b18cc765551a32f3ea500a0638d3a32410

tarcieri:
Thank you!

PR #636, opened 2021-09-14
Move format-related crates to RustCrypto/formats

tarcieri:
Splits out PKCS and X.509-related crates into a new repo:

https://github.com/RustCrypto/formats/pull/2

This PR leaves a directory for each of these crates in-place with a README.md pointing to the new location. We can eventually remove these, perhaps sometime after we've published new versions of all of the format-related crates which point to the new location.


newpavlov:
When do you plan to remove the folders with removal notices? Previously I was simply removing crates from source repositories and relied on new releases with updated `repository` fields.

tarcieri:
I'm cutting releases now with an updated `repository`, however I think it makes sense to leave the moved notifications around for a week or two

PR #643, opened 2021-09-30
Contributing iso-10126

### LINKED ISSUE #613: Contributing iso-10126
Dear maintainers, I wish to contribute the support for [iso-10126](https://en.wikipedia.org/wiki/Padding_(cryptography)#ISO_10126). It is basically a variation of `Pkcs7` or `AnsiX923` where arbitrary bytes are used to fill the buffer, except for the last byte. Would you accept such a contribution?

Context: ISO 10126 was withdrawn in 2007 but is still being used in some context, including in [XML enc](https://www.w3.org/TR/xmlenc-core1/#sec-Padding).

Considered alternatives for the "arbitrary padding bytes":
1. `0x00` like `AnsiX923`
2. last byte like `Pkcs7`
3. Something else to minimise confusion.

I am leaning towards (3), to help prevent confusion.
I am proposing `(len(buffer)+10126)%256`, such that it is deterministic yet users are less likely to confuse it with `PKCS#7` or `AnsiX923` or any other standard. On unpadding the ignored bytes would be ignored.

unbrice:
Fixes #613

### DIFF
--- a/block-padding/src/lib.rs
+++ b/block-padding/src/lib.rs
@@ -103,20 +103,6 @@ pub struct Pkcs7;
 
 impl Pkcs7 {
     #[inline]
-    fn pad<B: ArrayLength<u8>>(block: &mut Block<B>, pos: usize) {
-        // TODO: use bounds to check it at compile time
-        if B::USIZE > 255 {
-            panic!("block size is too big for PKCS#7");
-        }
-        if pos >= B::USIZE {
-            panic!("`pos` is bigger or equal to block size");
-        }
-        let n = (B::USIZE - pos) as u8;
-        for b in &mut block[pos..] {
-            *b = n;
-        }
-    }
-    #[inline]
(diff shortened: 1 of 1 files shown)


block-padding/src/lib.rs
```diff
@@ -116,9 +116,8 @@ impl<B: ArrayLength<u8>> Padding<B> for Pkcs7 {
             *b = n;
         }
     }
-
     #[inline]
-    fn unpad(block: &Block<B>) -> Result<&[u8], UnpadError> {
+    fn unpad<B: ArrayLength<u8>>(block: &Block<B>, strict: bool) -> Result<&[u8], UnpadError> {
```
tarcieri, line 120: So the `strict` flag is the main reason for changing this block to inherent methods?

If so, it only seems necessary for `unpad`, not `pad`.

newpavlov (approved):
Thank you!

PR #662, opened 2021-11-09
Document `#[zeroize(skip)]` in the zeroize crate

daxpedda:
This actually documents the `#[zeroize(skip)]` functionality in the zeroize crate, until now it was only documented in the `zeroize_derive` crate.

### DIFF
--- a/zeroize/src/lib.rs
+++ b/zeroize/src/lib.rs
@@ -69,6 +69,7 @@
 //! Additionally it supports the following attribute:
 //!
 //! - `#[zeroize(drop)]`: call `zeroize()` when this item is dropped
+//! - `#[zeroize(skip)]`: skips this field or variant when calling `zeroize()`
 //!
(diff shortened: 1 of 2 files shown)


zeroize/src/lib.rs
```diff
@@ -69,6 +69,7 @@
 //! Additionally it supports the following attribute:
 //!
 //! - `#[zeroize(drop)]`: call `zeroize()` when this item is dropped
+//! - `#[zeroize(skip)]`: skips this field or variant when calling `zeroize()`
```
tarcieri, line 72: It would probably be good to note that `drop` is a type-level attribute and `skip` is a field-level attribute

daxpedda: I separated them into different lists, I think it would go quiet well with #663. WDYT?

tarcieri:
Looks good, thanks

PR #663, opened 2021-11-10
Implement `#[zeroize(bound = "T: MyTrait")]`

daxpedda:
À la [`#[serde(bound = "T: MyTrait")]`](https://serde.rs/container-attrs.html#bound).

### DIFF
--- a/zeroize/derive/src/lib.rs
+++ b/zeroize/derive/src/lib.rs
@@ -6,8 +6,13 @@
 
 use proc_macro2::TokenStream;
 use quote::quote;
-use syn::{Attribute, Meta, NestedMeta};
-use synstructure::{decl_derive, BindStyle, BindingInfo, VariantInfo};
+use syn::{
+    parse::{Parse, ParseStream},
+    punctuated::Punctuated,
+    token::Comma,
+    Attribute, Lit, Meta, NestedMeta, Result, WherePredicate,
+};
+use synstructure::{decl_derive, AddBounds, BindStyle, BindingInfo, VariantInfo};
 
 decl_derive!(
     [Zeroize, attributes(zeroize)] =>
@@ -18,6 +23,8 @@ decl_derive!(
     ///
     /// On the item level:
     /// - `#[zeroize(drop)]`: call `zeroize()` when this item is dropped
+    /// - `#[zeroize(bound = "T: MyTrait")]`: this replaces any trait bounds
+    ///   inferred by zeroize-derive
     ///
     /// On the field level:
     /// - `#[zeroize(skip)]`: skips this field or variant when calling `zeroize()`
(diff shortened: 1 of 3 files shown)


tarcieri:
@daxpedda needs a rebase

daxpedda:
Done!

zeroize/derive/src/lib.rs
```diff
@@ -134,6 +160,49 @@ impl ZeroizeAttrs {
...
+                            }
+
+                            return;
+                        }
+                    }
+
+                    panic!(concat!(
+                        "The #[zeroize(bound)] attribute expects a name-value syntax with a string literal value.",
+                        "E.g. #[zeroize(bound = \"T: MyTrait\")]."
+                    ))
```
tarcieri, line 203: This much error handling really makes me wish we were using `proc-macro-error`, although I haven't figured out how to make it play nicely with `synstructure`.

tarcieri (approved):

zeroize/derive/src/lib.rs
daxpedda: I've tended to not use a lot of libraries when working with proc-macros exactly because of interoperability. I was close to suggesting to drop synstrucutre completely in favor of just using plain syn to skip parsing the user input and let rustc just say what's wrong with the where bounds.

I can look into it if you deem it necessary.

tarcieri: Yeah, I recently dropped `synstructure` from another proc macro crate, although `zeroize_derive` is better leveraging it.

`proc-macro-error` though I think is definitely worth it, since it allows emitting errors with diagnostic information that includes the original source spans, which makes error messages generated from proc macros substantially better.

tarcieri:
Will go ahead and merge this, although if you'd like to look into removing `synstructure` that'd be great.

zeroize/derive/src/lib.rs
daxpedda: I did it without proc-macro-error by just using [`to_compile_error`](https://docs.rs/syn/1.0.81/syn/parse/struct.Error.html#method.to_compile_error) until now. But our current strategy using panic isn't ideal, true enough.

daxpedda:
Will do!

zeroize/derive/src/lib.rs
tarcieri: I'm assuming you meant elsewhere outside the scope of this PR?

With `proc-macro-error` I've managed to make sure that practically every error case is tied to the input code that caused it.

daxpedda: > I'm assuming you meant elsewhere outside the scope of this PR?

Sry, lack of context indeed 😄.
> With `proc-macro-error` I've managed to make sure that practically every error case is tied to the input code that caused it.

That sounds great! I will take a look at it then.

PR #664, opened 2021-11-10
Improve `panic!` messages in `hex-literal`

thomascastleman:
Thanks for making the `hex-literal` crate! I was using it with an escaped hex string I'd copied from a C program, but ran into a panic because I'd accidentally left a semicolon after the strings I was calling `hex!()` on. It took me an embarrassing amount of time to discover the mistake, but I thought the panic messages might be able to provide a bit more info about what caused the failure.

This PR does only a few small things:
-  I added a bit more info to the panic messages when you encounter a non-literal, invalid character, or odd number of hex characters, to give you context about what's causing the issue. 
- I also added some internal documentation on some of the `TokenTreeIter` functions. 
- Lastly, I added some tests for the core `hex!()` macro functionality in the `tests/` directory (this was the only way I could find to set this up so the tests could invoke `hex!()`).

I didn't touch the changelog because I figure if this gets merged it's not a substantial enough change to warrant that. Thanks and I'd appreciate any feedback.

### DIFF
--- a/hex-literal/src/lib.rs
+++ b/hex-literal/src/lib.rs
@@ -86,12 +86,16 @@ struct TokenTreeIter {
 }
 
 impl TokenTreeIter {
+    /// Constructs a new `TokenTreeIter` from a given `proc_macro::Literal`.
+    ///
+    /// # Panics
+    /// This panics if the given `Literal` is not a string literal.
     fn new(input: Literal) -> Self {
         let mut buf: Vec<u8> = input.to_string().into();
 
         match buf.as_slice() {
             [b'"', .., b'"'] => (),
-            _ => panic!("expected string literals"),
+            _ => panic!("expected string literal, got `{}`", input),
         };
         buf.pop();
         let mut iter = buf.into_iter().exclude_comments();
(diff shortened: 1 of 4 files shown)


tarcieri (approved):
Looks good!

hex-literal/src/lib.rs
```diff
@@ -110,7 +119,7 @@ impl TokenTreeIter {
                 b'A'..=b'F' => v - 55,
                 b'a'..=b'f' => v - 87,
                 b' ' | b'\r' | b'\n' | b'\t' => continue,
-                _ => panic!("encountered invalid character"),
+                c => panic!("encountered invalid character: `{}`", c as char),
```
newpavlov, line 122: Note that it will print a wrong error message for non-ASCII characters. It's probably worth to change this line into something like this:
```rust
0..=127 => panic!("encountered invalid character: `{}`", c as char),
_ => panic!("encountered invalid non-ASCII character"),
```

newpavlov (commented):
Thank you, looks good! I have only one small comment.

Also can you please update the changelog with an `[UNRELEASED]` section?

hex-literal/src/lib.rs
thomascastleman: Good catch--thanks! Updated.

PR #665, opened 2021-11-11
`hex-literal`: ensure '/' as last byte causes error in `hex!()`

thomascastleman:
Hello again,

I noticed that, because the way comment exclusion works, having a single forward slash `/` as the last byte given to `hex!()` causes no error to be given where there should be one. For instance, the following program compiles without error:
```rust
let _: [u8; 0] = hex!("/");
```
Note that this doesn't happen if there are bytes after the `/` (in which a case it errors).

My fix adds a check in the implementation of `Iterator` for `ExcludingComments` for if there are no more bytes and we're in the `PotentialComment` state. In this case it simply yields the previous byte (which was the foward slash) so that `next_hex_val()` will try to parse it as hex and panic with "invalid character". This has the nice property that it's consistent with the error you get currently if a single slash is encountered not at the end of the stream.

### DIFF
--- a/hex-literal/CHANGELOG.md
+++ b/hex-literal/CHANGELOG.md
@@ -12,6 +12,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - More tests for `hex!()` macro
 - More internal documentation
 
+### Fixed
+- Make `hex!()` error when forward slash encountered as last byte
+
 ## 0.3.3 (2021-07-17)
 ### Added
 - Accept sequence of string literals ([#519])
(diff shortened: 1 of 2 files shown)


newpavlov:
Good catch! But can't we simply panic on `PotentialComment` without yielding the previous byte?

thomascastleman:
My thinking was to avoid replicating that same error message from `next_hex_val`, but if a panic right when we detect a final `/` is clearer, we can do that. How's this?

newpavlov:
Ah, `ExcludingComments` yields a single `/`, so it results in an "invalid character" error. I would change it a bit, but for now I think we can merge this PR as is.

PR #671, opened 2021-11-19
block-buffer v0.10

### LINKED ISSUE #8: Cover block-buffer with tests
Currently `block-buffer` has no tests and tested implicitly through its users.

### LINKED ISSUE #669: block-buffer: release 0.10 ?
Hi,
I was just wondering, `block-buffer` has been in `pre` (0.10.0-preX) state for almost a year now

In 0.10, this commit (https://github.com/RustCrypto/utils/pull/113) is of particular interest because of the "_make it almost panic-free_" part :
`This PR modifies block-buffer and block-padding APIs to reduce their surface and to make them mostly panic-free`

As quite some algorithms depends on it, releasing the version here and bumping dependencies should also make those panic-free too

For instance, with this mini-reproducer for the Shabal algorithm, still using block-buffer 0.9:

```
    let mut a = Shabal256::new();
    a.update(b"\x01");
    a.finalize();
```

When looking at the assembly:

```
  if ( (unsigned __int8)sub_9050(a1 + 8, 64LL, *(_QWORD *)a1) )
    sub_6E20(
      "called `Result::unwrap()` on an `Err` value  /home/test/.cargo/registry/src/github.com-1ecc6299db9ec823/shabal-0.3.0/src/shabal.rs",
      43LL,
      v20,
      &off_407B8,
      &off_407D8);
```

The related `unwrap` seems to be this one (https://github.com/RustCrypto/hashes/blob/master/shabal/src/shabal.rs#L248)

image

AFAIK, it should be panic-free in 0.10 and the compiler should optimize away this error handling

Again, sorry for the inconvenience and thanks for all the hard work on those different crates

newpavlov:
Split-off from #566

Notable changes:
- Introduced `Eager` and `Lazy` buffer type parameters which are responsible for defining cursor position invariant. For the `Eager` kind cursor position is always smaller than block size (i.e. buffer always contains at least one unfilled byte), while for `Lazy` it's less or equal (i.e. buffer can contain a complete block) The kinds are used as an important part of the `digest` core API.
- Removed dependency on `block-padding`. It results in a bit of code duplication in the hash crates, but reduces coupling between our crates.
- Block size is limited to 255 bytes on type level, it allows us to use `u8` for cursor position, which saves 7 bytes of state size on 64-bit targets and relaxes alignment requirements.

Right now it mostly tied to hash and MACs needs. I plan to experiment on cipher buffers separately inside the `cipher` crate. Cipher-related buffer functionality can be merged into this crate at a later date in separate versions.

Closes #8
Closes #669


block-buffer/src/lib.rs
```diff
@@ -1,48 +1,324 @@
...
+        // prevents panic elimination
+        let rem = self.size() - pos;
+        let n = input.len();
+        // Note that checking condition `pos + n < BlockSize` is
+        // equivalent to checking `n < rem`, where `rem` is equal
+        // to `BlockSize - pos`. Using the latter allows us to work
+        // around compiler accounting for possible overflow of
+        // `pos + n` which results in it inserting unreachable
+        // panic branches. Using `unreachable_unchecked` in `get_pos`
+        // we convince compiler that `BlockSize - pos` never underflows.
```
bridiver, line 129: I believe you mean overflow here, not underflow? If BlockSize - pos < 0, that's still overflow. Underflow only applies to floating point arithmetic.

tarcieri: https://cwe.mitre.org/data/definitions/191.html

> CWE-191: Integer Underflow (Wrap or Wraparound)
>
> The product subtracts one value from another, such that the result is less than the minimum allowable integer value, which produces a value that is not equal to the correct result.

bridiver: I guess it is saying specifically "integer underflow" and not "arithmetic underflow", but that's a bit of a confusing usage imo.
https://en.wikipedia.org/wiki/Arithmetic_underflow
> Storing values that are too low in an [integer](https://en.wikipedia.org/wiki/Integer_(computer_science)) variable (e.g., attempting to store −1 in an [unsigned](https://en.wikipedia.org/wiki/Signedness) integer) is properly referred to as [integer overflow](https://en.wikipedia.org/wiki/Integer_overflow), or more broadly, integer wraparound. The term underflow normally refers to floating point numbers only, which is a separate issue.

bridiver: We're debugging some a crash around here (actually in `get_pos()`) that only happens only on win x86 in release config. It started only recently so it may be related to rustc version and it's a bit odd because it appears to crash on https://github.com/RustCrypto/utils/pull/671/files#diff-dab38e6f4d05085a1ab6b7b064fb097fa39f1806d8b03dfbb80472aaa21851e0R171
I'll open a ticket for it if we can repro in a pure rust test

newpavlov: @bridiver 
If you do not trigger the debug assert on line 173 when you execute your program compiled in debug mode, then it's likely either a miscompilation (i.e. a compiler bug) or you have a memory corruption bug somewhere which damages block buffer memory.

Either way, reproduction will be appreciated. It may be worth to try inspect generated assembly in [Godbolt](https://rust.godbolt.org) (you can simply copy-paste `block-buffer` code into it with a bug reproduction code).

bridiver: > @bridiver If you do not trigger the debug assert on line 173 when you execute your program compiled in debug mode, then it's likely either a miscompilation (i.e. a compiler bug) or you have a memory corruption bug somewhere which damages block buffer memory.

There are no crashes in debug and this can be reliably reproduced across multiple builds by our tests. The tests passed in many previous releases and there are no changes to this particular code since the last successful run, but there are other changes like rustc version and possibly related changes to the build.

bridiver: Sorry, actually I take that back. I don't think we have tried a build that would trigger the debug asserts so we'll look at that as well.

bridiver: so we do see this with the debug assertions enabled
```
thread '<unnamed>' panicked at .\../../brave/third_party/rust/block_buffer/v0_10/crate/src/lib.rs:198:13:
assertion failed: false
```

bridiver: I still don't have a pure rust STR, but I'll open an issue

bridiver: btw - the panic message when we replicated this in rust was `attempt to subtract with overflow` :)

PR #675, opened 2021-11-23
Add inout crate and prepare block-padding v0.3

newpavlov:
Split-off from #566

This crate aims to generalize over in-place and buffer-to-buffer modes of operation and it's designed to be used in our cipher crates, i.e. block/stream ciphers, block modes, AE(AD)s.

### DIFF
--- a/inout/src/inout.rs
+++ b/inout/src/inout.rs
@@ -1,0 +1,174 @@
+use crate::InOutBuf;
+use core::{marker::PhantomData, ptr};
+use generic_array::{ArrayLength, GenericArray};
+
+/// Custom pointer type which contains one immutable (input) and one mutable
+/// (output) pointer, which are either equal or non-overlapping.
+pub struct InOut<'inp, 'out, T> {
+    pub(crate) in_ptr: *const T,
+    pub(crate) out_ptr: *mut T,
+    pub(crate) _pd: PhantomData<(&'inp T, &'out mut T)>,
+}
+
+impl<'inp, 'out, T> InOut<'inp, 'out, T> {
+    /// Reborrow `self`.
+    #[inline(always)]
+    pub fn reborrow<'a>(&'a mut self) -> InOut<'a, 'a, T> {
+        Self {
+            in_ptr: self.in_ptr,
+            out_ptr: self.out_ptr,
+            _pd: PhantomData,
+        }
+    }
+
+    /// Get immutable reference to the input value.
+    #[inline(always)]
+    pub fn get_in<'a>(&'a self) -> &'a T {
+        unsafe { &*self.in_ptr }
+    }
+
+    /// Get mutable reference to the output value.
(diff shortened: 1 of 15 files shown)


inout/src/inout_buf.rs
```diff
@@ -0,0 +1,357 @@
...
+    type IntoIter = InOutBufIter<'a, T>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        InOutBufIter { buf: self, pos: 0 }
+    }
+}
+
+impl<'a, T> InOutBuf<'a, T> {
+    /// Create `InOutBuf` from a single mutable reference.
+    pub fn from_mut(val: &mut T) -> Self {
```
steffahn, line 37: Missing lifetime `'a`!

newpavlov: Thank you!

inout/src/inout.rs
```diff
@@ -0,0 +1,149 @@
...
+    }
+}
+
+impl<'a, T, N: ArrayLength<T>> InOut<'a, GenericArray<T, N>> {
+    /// Returns `InOut` for the given position.
+    ///
+    /// # Panics
+    /// If `pos` greater or equal to array length.
+    pub fn get<'b>(&'b mut self, pos: usize) -> InOut<'b, T> {
+        assert!(pos <= N::USIZE);
```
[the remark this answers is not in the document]
newpavlov: Oh, great catch!

newpavlov:
@steffahn 
I have the following question, if you don't mind. Is it safe to define `get_in`/`get_out` methods on `InOut` like this?
```rust
pub fn get_in<'b>(&'b self) -> &'b T {
    unsafe { &*self.in_ptr }
}

 pub fn get_out<'b>(&'b mut self) -> &'b mut T {
    unsafe { &mut *self.out_ptr }
}
```

steffahn:
It's been a while since I last looked at this. It should probably be safe. 

To determine safety, there's two concerns:
* Does the code break any invariants for references/pointers, i.e. is is just plain always-UB? (E.g. aliasing mutable references, or invalidating pointers, etc...)
  * On this point, the methods seem very straightforward, so I don't see any danger
* Does the method offer a *sound* API? (I.e. are all the lifetimes right, is there some tricky way to abuse it, etc.) On this front, I think that remembering my proposed alternative-implementation helps. If `InOut<'a, T>` was just an `enum { Same(&'a mut T), Separate(&'a T, &'a mut T) }`, would the API be possible to implement in safe code? If that's true for all the API surface (perhaps ignoring things like `from_raw`), then it is for sure sound.
  * Trying this approach here [does work (at least for those two methods](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=cecafe124f1b1d3e82063992c50b6dcc), feel free to test if the remainder of the API does fit the same model), so all seems sound.

PR #692, opened 2021-12-21
block-buffer: Return Iterators instead of Block Slices

aewag:
This draft PR is meant to discuss a possible change in the `block-buffer` crate, which requires changes in the dependencies as well. The main advantage is to avoid the use of `unsafe`.

Currently, `block-buffer` works with raw pointers to transmute the incoming byte slices to block slices. I would like to propose returning an `Iterator` instead. To evaluate the possiblity, the `digest_block()` method is changed to work with iterators.

What do you think about this? Are there any limitations that are not considered?

#### Impact
I have not yet adapted other crates, which uses `block-buffer`. Thus, I don't have any numbers. If this seems like a valuable approach, I would setup a PoC adapting the `digest` and `sha2` crate to create some performance benchmarks. 

#### Open Points
Chaining Iterators would allow to call the `compress` function only once. Depending on the use-case, this could improve performance. As `ExactSizeIterators` and `chain` are not working together a different workaround is needed. Maybe working with `HintSize` can be a solution.

### DIFF
--- a/block-buffer/src/lib.rs
+++ b/block-buffer/src/lib.rs
@@ -113,7 +113,7 @@ where
     pub fn digest_blocks(
         &mut self,
         mut input: &[u8],
-        mut compress: impl FnMut(&[Block<BlockSize>]),
+        mut compress: impl FnMut(&mut BlockIterator<'_, BlockSize, Kind>),
     ) {
         let pos = self.get_pos();
         // using `self.remaining()` for some reason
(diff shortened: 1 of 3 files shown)


newpavlov:
I think it's an unnecessary complexity, which will ripple annoyingly through downstream crates. When const generics will be sufficiently advanced the removed `unsafe` parts will be simply replaced with the [`as_chunks`](https://doc.rust-lang.org/std/primitive.slice.html#method.as_chunks) method. Also note that in some cases we require continuity of the supplied blocks (e.g. when we supply blocks to an asm backend which iterates over blocks itself or with hashes like BLAKE3 which support parallel block processing), so the iterator approach will not work there.

aewag:
> I think it's an unnecessary complexity, which will ripple annoyingly through downstream crates. When const generics will be sufficiently advanced the removed `unsafe` parts will be simply replaced with the [`as_chunks`](https://doc.rust-lang.org/std/primitive.slice.html#method.as_chunks) method. 

Ah, that is something that I completely missed. I thought with `const-generics` there is still this problem.

> Also note that in some cases we require continuity of the supplied blocks (e.g. when we supply blocks to an asm backend which iterates over blocks itself or with hashes like BLAKE3 which support parallel block processing), so the iterator approach will not work there.

Missed as well.

Thank you for your feedback. I'll close this draft.

Issue #652, opened 2021-11-05
zeroize: add (back) `ZeroizeOnDrop` trait


tarcieri:
The `zeroize` crate used to have a `ZeroizeOnDrop` trait: https://github.com/iqlusioninc/crates/pull/168

It was removed in favor of `#[zeroize(drop)]` which largely turned out to be a mistake: https://github.com/iqlusioninc/crates/pull/188

Originally filed by @fjarri as https://github.com/iqlusioninc/crates/issues/757:

> Person A creates a crate with a type that has some secret data inside, say, `SecretKey { inner: SecretDataType }`. They implement `Zeroize` and `Drop` (or `#[zeroize(drop)]`)  for `SecretDataType`, and stop at that - the secret data is now safe (to the extend Rust can guarantee it), the `drop()` of `SecretKey` will call the zeroizing `drop()` of `SecretDataType`, so there's no need to define anything for `SecretKey` itself.
>
> Person B uses A's crate and wants to use `Secret<SecretKey>`. But they cannot, because `SecretKey` does not implement `Zeroize` - so they have to wrap it in a newtype and implement an empty `Zeroize` for it. And they still cannot assert in code that `SecretKey` keeps its secret - they have to rely on the docs. 
>
> So what are the responsibilities of A here? Should they define an (empty) `Zeroize` for `SecretKey`? Should they export a `Secret<Box<SecretKey>>`? What's the convention?
>
> I wonder if it would be more convenient to have a marker trait `ZeroizedOnDrop`, that A could implement for `SecretKey`, and `Secret` could require?

daxpedda:
Would a PR for this be welcome?

I would introduce a new derive called `ZeroizedOnDrop` that implements `Drop` without implementing `Zeroize`. Either derive `ZeroizedOnDrop` or `Zeroize(drop)` will implement the `ZeroizedOnDrop` marker trait too.

To get around the issue of requiring all fields to implement `Zeroize` I would abuse auto-deref:
```rust
trait AssertZeroizedOnDrop {
    fn zeroize_or_on_drop(&mut self);
}

impl<T: ZeroizedOnDrop> AssertZeroizedOnDrop for &mut T {
    fn zeroize_or_on_drop(&mut self) {}
}

trait AssertZeroize {
    fn zeroize_or_on_drop(&mut self);
}

impl<T: Zeroize> AssertZeroize for T {
    fn zeroize_or_on_drop(&mut self) {
        self.zeroize()
    }
}
```
[...]

PR #699, opened 2022-01-04
Implement `ZeroizeOnDrop`

daxpedda:
As discussed in #652.

### DIFF
--- a/zeroize/derive/src/lib.rs
+++ b/zeroize/derive/src/lib.rs
@@ -22,13 +22,25 @@ decl_derive!(
     /// Supports the following attributes:
     ///
     /// On the item level:
-    /// - `#[zeroize(drop)]`: call `zeroize()` when this item is dropped
+    /// - `#[zeroize(drop)]`: *deprecated* use `ZeroizeOnDrop` instead
     /// - `#[zeroize(bound = "T: MyTrait")]`: this replaces any trait bounds
     ///   inferred by zeroize-derive
     ///
     /// On the field level:
     /// - `#[zeroize(skip)]`: skips this field or variant when calling `zeroize()`
     derive_zeroize
+);
+
+decl_derive!(
+    [ZeroizeOnDrop, attributes(zeroize)] =>
+
+    /// Derive the `ZeroizeOnDrop` trait.
+    ///
+    /// Supports the following attributes:
+    ///
+    /// On the field level:
+    /// - `#[zeroize(skip)]`: skips this field or variant when calling `zeroize()`
+    derive_zeroize_on_drop
 );
 
 /// Name of zeroize-related attributes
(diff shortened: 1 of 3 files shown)


zeroize/derive/src/lib.rs
```diff
@@ -31,6 +31,18 @@ decl_derive!(
...
+decl_derive!(
+    [ZeroizeOnDrop, attributes(zeroize)] =>
+
+    /// Derive the `ZeroizeOnDrop` trait.
+    ///
+    /// Supports the following attributes:
+    ///
+    /// On the field level:
+    /// - `#[zeroize(skip)]`: skips this field or variant when calling `zeroize()`
+    derive_zeroize_only_drop
```
tarcieri, line 43: Was this supposed to be `derive_zeroize_on_drop`? I'm not sure what `only` is supposed to mean in this context

zeroize/derive/src/lib.rs
```diff
@@ -22,7 +22,7 @@ decl_derive!(
     /// Supports the following attributes:
     ///
     /// On the item level:
-    /// - `#[zeroize(drop)]`: call `zeroize()` when this item is dropped
+    /// - `#[zeroize(drop)]`: *deprecated* use `ZeroizeOnDrop` instead
```
tarcieri, line 25: To properly deprecate this I think it'd be good to emit a `#[deprecated(...)]` attribute in the generated code, although I'm fine with addressing that in a follow-up PR.

zeroize/derive/src/lib.rs
```diff
@@ -277,16 +315,25 @@ fn derive_zeroize_with_drop(s: synstructure::Structure<'_>) -> TokenStream {
...
         #zeroize_impl
 
         #[doc(hidden)]
         #drop_impl
+
+        #zeroize_on_drop_impl
     }
 }
 
+fn derive_zeroize_on_drop(s: &synstructure::Structure<'_>) -> TokenStream {
```
tarcieri, line 332: Aah okay, so `derive_zeroize_on_drop` does exist, but only adds the marker trait.

I think it would make more sense to also derive the `Drop` impl here?

daxpedda: Can't be done that way because the method body is completely different.

zeroize/derive/src/lib.rs
daxpedda: I am looking into it.

zeroize/derive/src/lib.rs
tarcieri: Switching the names sounds good to me.

daxpedda: Switching the names was quiet confusing, I changed it up a little, how does it look to you now?

tarcieri (approved):
This looks fine to me now.

If you'd like to add the `#[deprecated]` attribute in this PR go ahead, or otherwise I can just merge.

daxpedda:
Just checked it out, I can't get a warning to actually show up, feel free to merge, this can be done in a separate PR anyway. 

Do you know how this is done by any chance? I just emitted `#[deprecated = "..."]` above the trait implementation, did nothing.

daxpedda:
So https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-deprecated-attribute says:
> It cannot be applied to trait implementation items.

This is the related issue: https://github.com/rust-lang/rust/issues/39935.

tarcieri:
Yeah, you'd need to make some sort of pseudo-function (possibly in the zeroize crate) and call it from the deprecated impls.

Probably best to save something that convoluted for a follow-up PR.

tarcieri:
Thank you!

PR #700, opened 2022-01-04
`ZeroizeOnDrop` auto-deref

### LINKED ISSUE #652: zeroize: add (back) `ZeroizeOnDrop` trait

daxpedda:
This changes nothing from the users perspective, but will enable to implement `ZeroizeOnDrop` for fields that implement it replacing the need to `skip` them instead.

This will also validate that these fields types implement `ZeroizeOnDrop`.

See #652 for further discussion.
Builds on top of #699.

Fixes #652.

### DIFF
--- a/zeroize/src/lib.rs
+++ b/zeroize/src/lib.rs
@@ -263,6 +263,31 @@ pub trait Zeroize {
 #[allow(drop_bounds)]
 pub trait ZeroizeOnDrop: Drop {}
 
+#[doc(hidden)]
+pub mod __internal {
+    use super::*;
+
+    /// Auto-deref workaround for deriving `ZeroizeOnDrop`.
+    pub trait AssertZeroizeOnDrop {
+        fn zeroize_or_on_drop(&mut self);
+    }
+
+    impl<T: ZeroizeOnDrop> AssertZeroizeOnDrop for &mut T {
+        fn zeroize_or_on_drop(&mut self) {}
+    }
+
+    /// Auto-deref workaround for deriving `ZeroizeOnDrop`.
+    pub trait AssertZeroize {
+        fn zeroize_or_on_drop(&mut self);
+    }
+
+    impl<T: Zeroize> AssertZeroize for T {
+        fn zeroize_or_on_drop(&mut self) {
+            self.zeroize()
+        }
+    }
+}
+
 /// Marker trait for types whose `Default` is the desired zeroization result
 pub trait DefaultIsZeroes: Copy + Default + Sized {}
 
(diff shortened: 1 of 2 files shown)


zeroize/src/lib.rs
```diff
@@ -263,6 +263,30 @@ pub trait Zeroize {
 #[allow(drop_bounds)]
 pub trait ZeroizeOnDrop: Drop {}
 
+/// Auto-deref workaround for deriving `ZeroizeOnDrop`.
+#[doc(hidden)]
+pub trait AssertZeroizeOnDrop {
+    fn zeroize_or_on_drop(&mut self);
+}
+
+#[doc(hidden)]
```
tarcieri, line 272: I’d suggest making these proper sealed traits by placing them in a private module, rather than merely hiding them from the docs. Otherwise they’re still a part of the public API.

daxpedda: They need to be part of the public API, otherwise the derive proc-macro can't access them. Some alternatives:
- Rename them further to `__AssertZeroizeOnDrop`.
- Remove them from `zeroize` and let the proc-macro insert them. This has the disadvantage of requiring every `impl` block to add them, which might increase compile time.
- Mark traits as `unsafe`.

Or a combination of these.

tarcieri: > otherwise the derive proc-macro can't access them

Aah!

> Rename them further to __AssertZeroizeOnDrop.

Another option would be to put them in a module like `__internal`, which would also let you do `#[doc(hidden)]` in a single place.

> Mark traits as unsafe.

Since the trait doesn't deal strictly with memory safety I'd prefer not to do this.

daxpedda: Done!

tarcieri:
Thanks!

PR #709, opened 2022-01-13
zeroize: Fix docs

aewag:
I've experimented with zeroize v1.5.0-pre and just stumbled upon some minor doc typos.

### DIFF
--- a/zeroize/src/lib.rs
+++ b/zeroize/src/lib.rs
@@ -86,10 +86,10 @@
 //! ```
 //! # #[cfg(feature = "derive")]
 //! # {
-//! use zeroize::{Zeroize, ZeroizeDrop};
+//! use zeroize::{Zeroize, ZeroizeOnDrop};
 //!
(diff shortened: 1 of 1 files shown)


tarcieri:
Thanks! Curious why these weren't caught by doctests...

aewag:
Yeah, I also didn't see initially why this wasn't caught. Tests were executed so I assumed that's fine.

Just double checked and found the issue: The naming of the feature is `zeroize_derive` not `derive`. Thus, the scope was never enabled. I just tested it on my fork:

By only fixing the feature naming:
https://github.com/aewag/utils/pull/2/commits/f2bab38313dfc9d6d3ce2a7c473d9ac45b7b8d6b
The doctests fail as expected:
https://github.com/aewag/utils/runs/4806210245?check_suite_focus=true

Just created a 2nd PR that sits on top of the current master with these fixes applied. This ran successfully the tests on my machine
https://github.com/RustCrypto/utils/pull/710

PR #719, opened 2022-01-27
Fix double `mut` on `AssertZeroizeOnDrop`

daxpedda:
I found that this doesn't work. I hope this is not considered a breaking change?

### DIFF
--- a/zeroize/tests/zeroize_derive.rs
+++ b/zeroize/tests/zeroize_derive.rs
@@ -240,4 +240,19 @@ mod custom_derive_tests {
 
         assert_eq!(value.0, 0);
     }
+
+    #[test]
+    fn derive_only_zeroize_on_drop() {
+        #[derive(ZeroizeOnDrop)]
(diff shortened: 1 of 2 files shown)


zeroize/tests/zeroize_derive.rs
```diff
@@ -240,4 +240,19 @@ mod custom_derive_tests {
...
+        struct X([u8; 3]);
+
+        #[derive(ZeroizeOnDrop)]
+        struct Z(X);
+
+        let mut value = Z(X([1, 2, 3]));
+        unsafe {
+            std::ptr::drop_in_place(&mut value);
+        }
+        assert_eq!(&value.0 .0, &[0, 0, 0])
```
daxpedda, line 256: This describes the issue that is not compilable without this fix.

tarcieri:
Can merge this, yank `zeroize` v1.5.0 due to the bug, and hope for the best regarding compatibility of existing code

PR #727, opened 2022-02-01
DropTests - alternative tests on zeroize_derive using drop directly

mooreniemi:
Hello, I am trying to confirm the below tests are valid. I was surprised that when running `cargo test --all-features` (to turn on the `zeroize_derive` feature tests) and run these that `check_actual_memory_vec` fails but `check_actual_memory_string` succeeds.

```
running 15 tests
test custom_derive_tests::check_actual_memory_string ... ok
test custom_derive_tests::derive_bound ... ok
test custom_derive_tests::derive_enum_drop ... ok
test custom_derive_tests::check_actual_memory_vec ... FAILED
test custom_derive_tests::derive_enum_only_drop ... ok
test custom_derive_tests::derive_enum_skip ... ok
test custom_derive_tests::derive_enum_test ... ok
test custom_derive_tests::derive_inherit_both ... ok
test custom_derive_tests::derive_inherit_from_both ... ok
test custom_derive_tests::derive_inherit_zeroize_on_drop ... ok
test custom_derive_tests::derive_struct_drop ... ok
test custom_derive_tests::derive_struct_only_drop ... ok
test custom_derive_tests::derive_struct_skip ... ok
test custom_derive_tests::derive_struct_test ... ok
test custom_derive_tests::derive_tuple_struct_test ... ok

failures:

---- custom_derive_tests::check_actual_memory_vec stdout ----
thread 'custom_derive_tests::check_actual_memory_vec' panicked at 'assertion failed: `(left == right)`
  left: `[1, 2, 3]`,
 right: `[0, 0, 0]`', zeroize/tests/zeroize_derive.rs:308:9


failures:
    custom_derive_tests::check_actual_memory_vec

test result: FAILED. 14 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

error: test failed, to rerun pass '--test zeroize_derive'
```

Perhaps I am misunderstanding how `zeroize` should work on `drop` in this situation? My understanding is that`slice` gives me a view of the memory and will reflect underlying changes to it, so if I call `drop(value)` I should expect the `slice` view to show zero'd out memory.

### DIFF
--- a/zeroize/tests/zeroize_derive.rs
+++ b/zeroize/tests/zeroize_derive.rs
@@ -285,4 +285,49 @@ mod custom_derive_tests {
         }
         assert_eq!(&value.0 .0, &[0, 0, 0])
     }
+
+    #[test]
+    fn check_actual_memory_vec() {
+        #[derive(Zeroize, ZeroizeOnDrop)]
+        struct X([u8; 3]);
+
+        #[derive(Zeroize, ZeroizeOnDrop)]
+        struct Z(X);
+
+        let mut value = Z(X([1, 2, 3]));
+        let actual_memory_content = unsafe {
+            let b = value.0 .0;
+            std::slice::from_raw_parts(b.as_ptr(), b.len())
+        };
+
+        assert_eq!(actual_memory_content, value.0 .0);
+        assert_eq!(&actual_memory_content, &[1, 2, 3]);
+
+        drop(value);
+
+        assert_eq!(&actual_memory_content, &[0, 0, 0])
+    }
+
+    #[test]
+    fn check_actual_memory_string() {
+        const content: &str = "a string to make private";
+
+        #[derive(Zeroize, ZeroizeOnDrop)]
+        struct X(String);
+
+        let mut value: X = X(content.to_string());
+        let actual_memory_content = unsafe {
+            let b = value.0.as_bytes();
+            std::slice::from_raw_parts(b.as_ptr(), b.len())
+        };
+
+        assert_eq!(actual_memory_content, content.as_bytes());
+
+        drop(value);
+
+        assert_eq!(
+            &actual_memory_content,
+            &[0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
+        )
+    }
 }
--- a/zeroize/tests/zeroize_derive.rs
+++ b/zeroize/tests/zeroize_derive.rs
@@ -287,7 +287,7 @@ mod custom_derive_tests {
     }
 
     #[test]
-    fn check_actual_memory_vec() {
+    fn check_actual_memory_array() {
         #[derive(Zeroize, ZeroizeOnDrop)]
         struct X([u8; 3]);
 


zeroize/tests/zeroize_derive.rs
```diff
@@ -285,4 +285,49 @@ mod custom_derive_tests {
...
+        let mut value = Z(X([1, 2, 3]));
+        let actual_memory_content = unsafe {
+            let b = value.0 .0;
+            std::slice::from_raw_parts(b.as_ptr(), b.len())
+        };
+
+        assert_eq!(actual_memory_content, value.0 .0);
+        assert_eq!(&actual_memory_content, &[1, 2, 3]);
+
+        drop(value);
```
mooreniemi, line 306: Is the issue that `drop(value)` does not drop `value.0 .0`, but can only drop `value.0` (as in the below `String` example)?

tarcieri:
cc @daxpedda

zeroize/tests/zeroize_derive.rs
```diff
@@ -285,4 +285,49 @@ mod custom_derive_tests {
         }
         assert_eq!(&value.0 .0, &[0, 0, 0])
     }
+
+    #[test]
+    fn check_actual_memory_vec() {
```
tarcieri, line 290: Nit: there's no `Vec` involved here. Perhaps:

```suggestion
    fn check_actual_memory_array() {
```

zeroize/tests/zeroize_derive.rs
```diff
@@ -285,4 +285,49 @@ mod custom_derive_tests {
...
+            let b = value.0 .0;
+            std::slice::from_raw_parts(b.as_ptr(), b.len())
+        };
+
+        assert_eq!(actual_memory_content, value.0 .0);
+        assert_eq!(&actual_memory_content, &[1, 2, 3]);
+
+        drop(value);
+
+        assert_eq!(&actual_memory_content, &[0, 0, 0])
```
tarcieri, line 308: This is a use-after-free and there are no guarantees the value will actually be zero. You can do this instead:

```suggestion
        assert_ne!(&actual_memory_content, &[1, 2, 3])
```

You might try using a larger array to reduce the possibility that it's overwritten with something that might just so happen to be the expected pattern to be cleared.

tarcieri:
My best guess as to an explanation is that Rust's move semantics are making a copy, since it's a value that lives on the stack and you are passing it by-value to `drop`.

In general you can't expect this pattern to work with moved values. If you want to ensure a value that lives on the stack is properly zeroized, it can never be moved. See:

https://docs.rs/zeroize/latest/zeroize/#stackheap-zeroing-notes

zeroize/tests/zeroize_derive.rs
mooreniemi: I thought, maybe because I reused `let mut value = Z(X([1, 2, 3]));` and it's just coincidence? But switching to `[4, 5, 8]` (`git grep` says this is unique across the crate) yields same result:

```
failures:

---- custom_derive_tests::check_actual_memory_vec stdout ----
thread 'custom_derive_tests::check_actual_memory_vec' panicked at 'assertion failed: `(left != right)`
  left: `[4, 5, 8]`,
 right: `[4, 5, 8]`', zeroize/tests/zeroize_derive.rs:308:9


failures:
    custom_derive_tests::check_actual_memory_vec

test result: FAILED. 14 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

error: test failed, to rerun pass '--test zeroize_derive'
```

tarcieri: The problem I'm trying to highlight here is the use-after-free.

You are storing a pointer to potentially dropped memory and trying to read its value.

All bets are off after that. It can be arbitrarily overwritten.

My point is your 3-byte sequence is small enough it's possible that it just-so-happens it is overwritten with that value. I'm not saying that's actually what's happening here: see my explanation below. But to write a reliable test you need to lower the probability that it isn't being overwritten by a value which just so happens to match the original pattern.

mooreniemi: Yes, I follow you, but even with increasingly unlikely coincidences I still see the same failure, which is why I am inclined to think something else is happening and drew my comparison about `.0 .0` and `.0`. Re "see my explanation below" I am not seeing alternative explanation comment on this page. I'll check Conversation tab.

```
---- custom_derive_tests::check_actual_memory_vec stdout ----
thread 'custom_derive_tests::check_actual_memory_vec' panicked at 'assertion failed: `(left != right)`
  left: `[4, 5, 8, 23, 44, 101, 44, 0, 39]`,
 right: `[4, 5, 8, 23, 44, 101, 44, 0, 39]`', zeroize/tests/zeroize_derive.rs:308:9
```

tarcieri: Try replacing `drop` with a scope and see what happens.

mooreniemi:
Ok then essentially this test is _not_ valid for both types, but only the wrapped `String` (perhaps just by luck). Adding `Pin` into the mix makes sense from what you're saying. I think this answers my q and we can close.

daxpedda:
I usually use [`std::ptr::drop_in_place()`](https://doc.rust-lang.org/core/ptr/fn.drop_in_place.html) to avoid use-after-free.

PR #730, opened 2022-02-06
[WIP] crypto-intrinsics

tarcieri:
With the forthcoming stabilization of inline assembly, I thought we could start putting together a crate which provides wrappers for assembly instructions which are useful for cryptography but lack proper `core::arch` wrappers.

More importantly, using inline assembly allows us to provide a sort of black box which LLVM will not interfere with, which is problematic using anything else besides ASM. For example, it's otherwise not possible to correctly emit CMOV instructions on x86 platforms with LLVM because the `x86-cmov-conversion` pass which will rewrite them as branches. For more details, see:

https://dsprenkels.com/cmov-conversion.html


tarcieri:
ADX and MULX are also good candidates, and particularly problematic to emit outside of inline assembly due to LLVM bugs:

https://github.com/rust-lang/stdarch/issues/666

crypto-intrinsics/src/x86_64.rs
```diff
@@ -0,0 +1,62 @@
...
+
+use core::arch::asm;
+
+/// Move if zero.
+///
+/// Uses a `test` instruction to check if the given `condition` value is
+/// equal to zero, then calls `cmovz` (a.k.a. `cmove`) to conditionally move
+/// `src` to `dst` when `condition` is equal to zero.
+#[inline]
+pub fn cmovz(condition: u64, src: u64, dst: &mut u64) {
```
newpavlov, line 11: I think you can use `#[inline(always)]` on such intrinsics.

newpavlov (commented):
Interesting. Is it feasible for those intrinsics to be added to `core::arch` eventually? Also I wonder about fragility of flags used with ADX/MULX. If we are to chain several calls to such intrinsics, are we sure that compiler can not reorder some intrinsics clobbering the flags between those calls?

crypto-intrinsics/src/x86_64.rs
tarcieri: Added

tarcieri:
>  Is it feasible for those intrinsics to be added to core::arch eventually?

I think the only way would be if they used inline assembly themselves. [In fact I requested as much](https://github.com/rust-lang/stdarch/issues/666#issuecomment-953271708) and they seemed skeptical.

> If we are to chain several calls to such intrinsics, are we sure that compiler can not reorder some intrinsics clobbering the flags between those calls?

I think it still relies on careful inspection of the generated assembly, but really the important thing IMO is that LLVM does not insert additional instructions that clobber the flags, as otherwise I think the data dependencies should take care of reordering which would affect the flags.

crypto-intrinsics/src/x86_64.rs
```diff
@@ -0,0 +1,148 @@
...
+/// with carry operations and multiplications.
+///
+/// If the first and second operand are identical, it will contain the high
+/// half of the multiplication result.
+///
+/// # Safety
+///
+/// This function requires support for the Intel BMI2 (Bit Manipulation
+/// Instruction Set 2) extension to the x86 instruction set.
+#[target_feature(enable = "bmi2")]
```
tarcieri, line 116: I was able to add this for MULX, as it requires BMI2.

I tried to do the same thing with ADX, but unfortunately:

```
error[E0658]: the target feature `adx` is currently unstable
  --> crypto-intrinsics/src/x86_64.rs:26:18
   |
26 | #[target_feature(enable = "adx")]
   |                  ^^^^^^^^^^^^^^
```

dbl/src/lib.rs
```diff
@@ -29,12 +29,14 @@ pub trait Dbl {
...
     /// polynomial with the minimal number of ones.
+    #[must_use]
     fn dbl(self) -> Self;
 
     /// Reverse double block. (alternatively: divbide block by x)
     ///
     /// If least significant bit of the block equals to zero will return
     /// `block>>1`, otherwise `(block>>1)^(1<<n)^(C>>1)`
+    #[must_use]
     fn inv_dbl(self) -> Self;
```
tarcieri, line 40: These were needed for clippy to succeed after bumping MSRV. I can submit them in a separate PR.

newpavlov: Having those changes in this PR is fine by me.

crypto-intrinsics/src/x86_64.rs
tarcieri: It looks like a PR to stabilize ADX stalled some time in 2019: https://github.com/rust-lang/rust/pull/60109

tarcieri:
Removed draft/WIP and slimmed the initial implementation down to just `cmovz` and `cmovnz` on `x86` and `x86_64` targets.

Namely I removed all ADX/MULX support as we need a solution which supports carry chains emitted as a single `asm!` block in order to ensure flags don't get clobbered by LLVM (which is the whole point of using ADX/MULX at all).

Potentially in the future we can add some more of the CMOVcc family, and try to come up with something like a macro solution for ADX/MULX chains which emits an `asm!` macro with a fully formed set of assembly instructions that comprise the entire carry chain.

If it winds up this crate only implements CMOV though, I've gone ahead and snagged the `cmov` crate as well, and we could potentially rename this crate to that.

newpavlov:
Can you update the CI workflow with cargo cache and minimal versions job? The latter is less needed since the crate does no have any dependencies, but I guess it's better to keep workflows consistent with each other.

crypto-intrinsics/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+"""
+version = "0.0.0"
+authors = ["RustCrypto Developers"]
+license = "Apache-2.0 OR MIT"
+repository = "https://github.com/RustCrypto/utils/tree/master/crypto-intrinsics"
+categories = ["cryptography"]
+keywords = ["crypto", "cmov", "intrinsics"]
+readme = "README.md"
+edition = "2018" # Can't bump to 2021 due to pre-1.56 MSRV crates in the same workspace
+rust-version = "1.59"
```
newpavlov, line 16: Wouldn't it be better to use 2021 edition and the separate workspace trick for now?

PR #737, opened 2022-02-18
Test and fix min versions build

newpavlov:


zeroize/derive/CHANGELOG.md
```diff
@@ -4,7 +4,13 @@ All notable changes to this project will be documented in this file.
...
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## 1.3.1 (2021-01-14)
+## 1.3.2 (2022-02-22)
+### Fixed
+- Min versions build ([#732])
+
+[#732]: https://github.com/RustCrypto/utils/pull/732
+
+## 1.3.1 (2021-01-14) [YANKED]
```
newpavlov, line 13: @tarcieri 
Note that I plan to yank 1.3.1. Without doing so it will be very annoying to fix min versions build for crates which depend on `zeroize_derive` through `zeroize`.

PR #741, opened 2022-02-26
Initial `cmov` crate

tarcieri:
Implements safe wrappers for the `cmovz` and `cmovnz` instructions which work on `x86` and `x86_64`, implemented using inline assembly.

Using inline assembly allows us to provide a sort of black box which LLVM will not interfere with, which is otherwise problematic when using anything besides ASM.

It's otherwise not possible to correctly emit CMOV instructions on x86 platforms with LLVM because the `x86-cmov-conversion` pass which will potentially rewrite them as branches.

For more details, see:

https://dsprenkels.com/cmov-conversion.html

### DIFF
--- a/cmov/README.md
+++ b/cmov/README.md
@@ -1,0 +1,91 @@
+# [RustCrypto]: Conditional Move Intrinsics
+
+[![Crate][crate-image]][crate-link]
+[![Docs][docs-image]][docs-link]
+![Apache 2.0/MIT Licensed][license-image]
+![MSRV][rustc-image]
+[![Build Status][build-image]][build-link]
+
+Conditional move CPU intrinsics which are guaranteed to execute in
+constant-time and not be rewritten as branches by the compiler.
+
+Provides wrappers for the [CMOV family] of instructions on x86/x86_64 CPUs.
+
+[Documentation][docs-link]
+
+## About
+
+Conditional move intrinsics are a form of [predication] which allows selection
+of one or more values without using branch instructions, thus making the
+selection constant-time with respect to the values, and not subject to CPU
+execution features which might introduce timing or other microarchitectural
+sidechannels introduced by branch prediction or other speculative execution
+features.
+
+Intel has confirmed that all extant CPUs implement the CMOV family of
+instructions in constant-time, and that this property will hold for future
+Intel CPUs as well.
+
+This crate provides wrappers for the CMOV instructions implemented in terms
+of inline assembly as stabilized in Rust 1.59. This means the implementation
+is a black box that will not be rewritten by e.g. LLVM's architecture-specific
+lowerings, such as the [x86-cmov-conversion] pass.
+
+## Supported target architectures
+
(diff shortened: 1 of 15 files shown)


cmov/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+This crate provides wrappers for the CMOV instructions implemented in terms
+of inline assembly as stabilized in Rust 1.59. This means the implementation
+is a black box that will not be rewritten by e.g. LLVM's architecture-specific
+lowerings, such as the [x86-cmov-conversion] pass.
+
+## Supported target architectures
+
+This crate will compile on any Rust target, however the intrinsics are only
+made available on specific target architectures listed below:
```
tarcieri, line 37: Alternatively I could make it `compile_error!` on unsupported architectures. That might make more sense.

cmov/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+- [x] `x86`
+- [x] `x86_64`
+- [ ] `aarch64`
+- [ ] Others?
```
tarcieri, line 42: It looks like there are similar instructions which might provide equivalent semantics on various other architectures:

https://en.wikipedia.org/wiki/Load-link/store-conditional

cmov/README.md
tarcieri: Updated the crate to use the `compile_error!` approach for now

cmov/src/lib.rs
```diff
@@ -0,0 +1,111 @@
...
+///
+/// # Returns
+/// - `condition` is zero: `0`
+/// - `condition` is non-zero: `1`
+#[cfg(not(any(target_arch = "x86", target_arch = "x86_64")))]
+#[inline(always)]
+fn is_non_zero(condition: usize) -> usize {
+    const SHIFT_BITS: usize = core::mem::size_of::<usize>() - 1;
+    ((condition | (!condition).wrapping_add(1)) >> SHIFT_BITS) & 1
+}
```
tarcieri, line 88: This can probably be improved, but seems to do the basic trick.

On x86, it emits the following assembly:

https://godbolt.org/z/eccaGzGsv

```asm
example::cmovz:
        mov     rax, rdi
        neg     rax
        or      rax, rdi
        shr     rax, 7
        mov     ecx, eax
        and     ecx, 1
        or      rax, -2
        neg     rcx
        and     rcx, qword ptr [rdx]
        add     rax, 1
        and     rax, rsi
        or      rax, rcx
        mov     qword ptr [rdx], rax
        ret

example::cmovnz:
        mov     eax, edi
        neg     eax
        or      eax, edi
        shr     eax, 7
        and     eax, 1
        mov     rcx, rax
        add     rax, -1
        and     rax, qword ptr [rdx]
        neg     rcx
        and     rcx, rsi
        or      rcx, rax
        mov     qword ptr [rdx], rcx
        ret
```

We could potentially lint that the generated assembly uses instructions from a specific allow list on the platforms cross tests run on.

PR #744, opened 2022-03-01
Add cmovz/cmovnz support for aarch64.

codahale:
This passes tests and I _believe_ it works, but I also have the assembly programming experience of your average kindergartner, so a second opinion would be worth getting.

Running `cargo rustc --release -- --emit asm` on the portable `cmovz` yields this:

```
        neg     w8, w0
        orr     w8, w8, w0
        ubfx    x8, x8, #7, #1
        ldr     x9, [x2]
        neg     x10, x8
        and     x9, x9, x10
        sub     x8, x8, #1
        and     x8, x8, x1
        orr     x8, x9, x8
        str     x8, [x2]
        ret
```

And a naive, branching version of `cmovz` yields this:

```
        cbz     x0, LBB1_2
        ret
LBB1_2:
        str     x1, [x2]
        ret
```

Which makes me think `cmp`/`csel` is the right choice.

### DIFF
--- a/cmov/Cargo.toml
+++ b/cmov/Cargo.toml
@@ -3,7 +3,7 @@ name = "cmov"
 description = """
 Conditional move CPU intrinsics which are guaranteed to execute in
 constant-time and not be rewritten as branches by the compiler.
-Provides wrappers for the CMOV family of instructions on x86/x86_64 CPUs.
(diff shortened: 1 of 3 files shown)


tarcieri:
Looks good, thank you!

Can you update the documentation to reflect AArch64 assembly is available as well?

tarcieri:
Thank you!

PR #746, opened 2022-03-02
cmov: Set asm options

xfix:
Those options tell the compiler that ASM statements do not read or write memory, can be removed if the dst ends up unused during compilation, and that they don't push data to the stack.

This can help compiler generate more optimal code. See https://doc.rust-lang.org/nightly/reference/inline-assembly.html#options for documentation of those options.


tarcieri (approved):

tarcieri:
Thanks!

newpavlov:
Is it correct to use `nomem` here? The goal of the blocks is to literally write data to `dst` if condition is met. Same for `pure`, IIUC writing to a pointer makes blocks not pure.

PR #749, opened 2022-03-16
zeroize: add support for ARM64 SIMD registers (nightly-only)

tarcieri:
Similar to the implementation for `x86`/`x86_64`, this commit adds nightly-only support for zeroizing ARM64 SIMD registers.

Support is gated behind an `aarch64` feature so as to avoid breaking compilation on stable Rust. The feature is a no-op on non-`aarch64` targets.

### DIFF
--- a/.github/workflows/zeroize.yml
+++ b/.github/workflows/zeroize.yml
@@ -64,4 +64,26 @@ jobs:
           override: true
           profile: minimal
       - run: cargo test --release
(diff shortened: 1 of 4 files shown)


tarcieri:
This is blocked on https://github.com/rust-lang/rust/issues/95002

tarcieri:
Pinned to an earlier nightly. It's working for now.

PR #752, opened 2022-03-18
cpufeatures: Add support for AArch64 Android (essentially a Linux variant)

### LINKED ISSUE #751: cpufeatures: Use getauxval for aarch64 Android as well
`libc::getauxval` was added for Android in rust-lang/libc#1987, and is available on all aarch64 Android devices, so we should be able to detect features there as well.

jrose-signal:
Checking against [google/cpu_features](/google/cpu_features) it looks like Android follows Linux exactly in how it defines `getauxval`, so it really is just turning it on in aarch64.rs.

I have not actually tried *running* this on a 64-bit Android device yet; I'll get to that a bit later today. (Have to get that development environment set up again.) It does pass `cargo build --target aarch64-linux-android`, though (unsurprisingly).

Fixes #751.

### DIFF
--- a/cpufeatures/Cargo.toml
+++ b/cpufeatures/Cargo.toml
@@ -19,3 +19,6 @@ libc = "0.2.68"
 
 [target.'cfg(all(target_arch = "aarch64", target_os = "linux"))'.dependencies]
 libc = "0.2.68"
+
+[target.aarch64-linux-android.dependencies]
+libc = "0.2.68"
--- a/cpufeatures/src/aarch64.rs
+++ b/cpufeatures/src/aarch64.rs
@@ -21,7 +21,7 @@ macro_rules! __unless_target_features {
 }
 
 // Linux runtime detection of target CPU features using `getauxval`.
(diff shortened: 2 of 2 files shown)


cpufeatures/Cargo.toml
```diff
@@ -19,3 +19,6 @@ libc = "0.2.68"
 
 [target.'cfg(all(target_arch = "aarch64", target_os = "linux"))'.dependencies]
 libc = "0.2.68"
+
+[target.aarch64-linux-android.dependencies]
+libc = "0.2.68"
```
tarcieri, line 24: Kind of confused why the above doesn't suffice for this. Is `target_arch` or `target_os` actually different from the above on Android?

jrose-signal: Yeah, that's the only reason the existing implementation *didn't* work.

jrose-signal:
> I have not actually tried *running* this on a 64-bit Android device yet; I'll get to that a bit later today.

(Just checking in that I did, in fact, get to this, and it does work.)

PR #761, opened 2022-04-20
feat: make `Zeroizing` transparent for cheap conversions

npmccallum:
Sometimes libraries want to be generic across types like `Vec<u8>` and
`Box<[u8]>`. Therefore, they use bounds like `T: AsRef<[u8]>`. The
`Zeroizing<Vec<u8>>` type should be transparently equivalent to
`Vec<u8>` in this regard. This allows `Zeroizing` to be used with all
such bounds.

Signed-off-by: Nathaniel McCallum <nathaniel@profian.com>

### DIFF
--- a/zeroize/src/lib.rs
+++ b/zeroize/src/lib.rs
@@ -580,6 +580,26 @@ where
     }
 }
 
+impl<T, Z> AsRef<T> for Zeroizing<Z>
+where
+    T: ?Sized,
+    Z: AsRef<T> + Zeroize,
+{
+    fn as_ref(&self) -> &T {
+        self.0.as_ref()
+    }
+}
+
+impl<T, Z> AsMut<T> for Zeroizing<Z>
+where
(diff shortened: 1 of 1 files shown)


zeroize/src/lib.rs
```diff
@@ -580,6 +580,26 @@ where
     }
 }
 
+impl<T: ?Sized, Z> AsRef<T> for Zeroizing<Z>
+where
+    Z: Zeroize,
+    Z: AsRef<T>,
```
tarcieri, line 586: I think it's a bit weird to have some bounds inline and some in the where clause.

```suggestion
impl<T, Z> AsRef<T> for Zeroizing<Z>
where
    T: ?Sized,
    Z: Zeroize,
    Z: AsRef<T>,
```

npmccallum: Fixed.

PR #768, opened 2022-06-02
fiat-constify: initial crate

tarcieri:
This crate is a tool for transforming `fiat-crypto`-generated field arithmetic into `const fn` form.

It performs mechanical translation to the parsed AST of the code, making some assumptions about how it's structured and rewriting the usages of mutable references into a functional form.

Manual tests confirm the output from the p386 base field compiles.

### DIFF
--- a/.github/workflows/fiat-constify.yml
+++ b/.github/workflows/fiat-constify.yml
@@ -1,0 +1,37 @@
+name: fiat-constify
+
+on:
+  pull_request:
+    paths:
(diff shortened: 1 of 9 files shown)


tarcieri:
Still need to actually integrate the generated code into a crate (e.g. `p384`) and test what performance impact it has, if any.

FWIW here's an upstream issue about adding `const fn` support to `fiat-crypto` proper: https://github.com/mit-plv/fiat-crypto/issues/1086

PR #779, opened 2022-07-06
cpufeatures: Disable all target features under miri

### LINKED ISSUE #778: cpufeatures: requesting (minimal) miri support
[`miri`](https://github.com/rust-lang/miri) doesn't support inline assembly, including `cpuid`. But cpufeatures could treat it like it does SGX: if a target feature isn't on at compile time, it's not available at run time.

jrose-signal:
~~[miri](https://github.com/rust-lang/miri) is an interpreter and doesn't support inline assembly. Fortunately, we already have precedent for an x86 target without cpuid: SGX. Treat miri like SGX: only features that are enabled at compile time are available at run time.~~

[miri](https://github.com/rust-lang/miri) is an interpreter, and though it tries to emulate the target CPU it does not support any target features.

Fixes #778.

### DIFF
--- a/cpufeatures/src/lib.rs
+++ b/cpufeatures/src/lib.rs
@@ -59,12 +59,17 @@
     html_favicon_url = "https://raw.githubusercontent.com/RustCrypto/media/6ee8e381/logo.svg"
 )]
 
+#[cfg(not(miri))]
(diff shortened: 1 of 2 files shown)


tarcieri:
Looks good now, thanks!

PR #780, opened 2022-07-18
feat: add serde support

npmccallum:
This (optional) feature implements Serialize/Deserialize for Zeroizing when
the inner type implements these traits. This makes it possible to use
Zeroizing in serde contexts.

Signed-off-by: Nathaniel McCallum <nathaniel@profian.com>

### DIFF
--- a/zeroize/Cargo.toml
+++ b/zeroize/Cargo.toml
@@ -17,6 +17,7 @@ keywords = ["memory", "memset", "secure", "volatile", "zero"]
 edition = "2018"
 
 [dependencies]
+serde = { version = "1.0", default-features = false, optional = true }
 zeroize_derive = { version = "1.3", path = "derive", optional = true }
 
 [features]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -221,6 +221,7 @@ dependencies = [
 name = "zeroize"
 version = "1.5.6"
 dependencies = [
+ "serde",
  "zeroize_derive",
 ]
 
--- a/zeroize/src/lib.rs
+++ b/zeroize/src/lib.rs
@@ -715,6 +715,34 @@ where
     }
 }
 
+#[cfg(feature = "serde")]
+impl<Z> serde::Serialize for Zeroizing<Z>
+where
+    Z: Zeroize + serde::Serialize,
+{
+    #[inline(always)]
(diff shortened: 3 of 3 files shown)


zeroize/Cargo.toml
```diff
@@ -17,6 +17,7 @@ keywords = ["memory", "memset", "secure", "volatile", "zero"]
 edition = "2018"
 
 [dependencies]
+serde = { version = "1.0", optional = true }
```
daxpedda, line 20:
```suggestion
serde = { version = "1.0", default-features = false, optional = true }
```

daxpedda (changes requested):

zeroize/Cargo.toml
tarcieri: We generally try to enable the minimum number of features needed for a particular purpose.

Using `default-features = false` makes sense for both binaries and libraries, and is a requirement to use a prospective `serde` feature in `no_std` environments. See: https://serde.rs/no-std.html

tarcieri:
I'm curious if there's an attribute recipe we could come up with so as to avoid directly coupling `zeroize` to `serde`.

`#[serde(from = "T")]` should suffice for the deserializer. Offhand I'm not sure if there's a way to get `serde` to leverage traits like `AsRef`/`Borrow`/`Deref` to obtain a reference the inner type prior invoking `Serialize::serialize`, however.

See: https://serde.rs/container-attrs.html

zeroize/Cargo.toml
npmccallum: Done.

npmccallum:
@tarcieri I don't think that's desirable. `#[serde(from = "T")]` is for containers. So you'd have to do a newtype for every use. And even if we found an indirect way to leverage serde, it would probably have non-obvious side-effects.

An optional coupling between zeroize and serde is zero cost for everyone not using the feature. It is obvious and well-documented with no side-effects. That seems to me the best approach.

daxpedda:
@tarcieri are you saying we should use `#[serde(from = "T")]` on `Zeroizing` instead of doing a manual implementation? I don't understand how this avoids coupling `zeroize` to `serde` though.

npmccallum:
So can we merge this then?

PR #781, opened 2022-07-20
Add `const SIZE: usize` to `ArraySize` in hybrid-array

ryankurte:
Equivalent to `ArrayOps::SIZE` but useful where you have generics over the size instead of the array (like `digest::Digest`).


ryankurte:
ohh, `OutputSize::USIZE` should do the job already so this _shouldn't_ be required if i can work out why that's not playing ball...

tarcieri:
The naming here is admittedly not great. I will look into potentially cleaning this up in the next breaking release.

PR #787, opened 2022-08-01
Fix deriving `ZeroizeOnDrop` on items with generics

### LINKED ISSUE #786: `ZeroizeOnDrop` derivation does not work for a `Box<T>` field
The code 
```rust
#[derive(ZeroizeOnDrop)]
pub struct SecretBox<T: Zeroize>(Box<T>);
```
leads to the following compilation error (rustc 1.61):
```
error[E0367]: `Drop` impl requires `T: Drop` but the struct it is implemented for does not
 --> src/main.rs:3:10
  |
3 | #[derive(ZeroizeOnDrop)]
  |          ^^^^^^^^^^^^^
  |
note: the implementor must specify the same requirement
 --> src/main.rs:4:1
  |
4 | pub struct SecretBox<T: Zeroize>(Box<T>);
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  = note: this error originates in the derive macro `ZeroizeOnDrop` (in Nightly builds, run with -Z macro-backtrace for more info)
```

Meanwhile, a trivial drop implementation compiles fine:
```rust
impl<T: Zeroize> Drop for SecretBox<T>
{
    fn drop(&mut self) {
        self.0.zeroize()
    }
}

impl<T: Zeroize> ZeroizeOnDrop for SecretBox<T> {}
```
What can be the reason the automatic derivation does not work then? I am not sure how to interpret the compiler error.

daxpedda:
I don't actually know how to test this, because you can't access the dropped memory of a `Box` without a custom allocator I suppose (EDIT: same for `Vec`). So I just made sure it builds.

I'm also not a 100% sure my assumption is correct, but as far as I understand any `Drop` bounds are truly unnecessary in this case.

Fixes #786.

### DIFF
--- a/zeroize/derive/src/lib.rs
+++ b/zeroize/derive/src/lib.rs
@@ -71,7 +71,7 @@ fn derive_zeroize(mut s: synstructure::Structure<'_>) -> TokenStream {
 fn derive_zeroize_on_drop(mut s: synstructure::Structure<'_>) -> TokenStream {
     let zeroizers = generate_fields(&mut s, quote! { zeroize_or_on_drop });
 
-    let drop_impl = s.gen_impl(quote! {
+    let drop_impl = s.add_bounds(AddBounds::None).gen_impl(quote! {
         gen impl Drop for @Self {
             fn drop(&mut self) {
                 use zeroize::__internal::AssertZeroize;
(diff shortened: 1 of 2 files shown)


tarcieri:
Thanks!

PR #792, opened 2022-08-16
cpufeatures: check AVX availability when detecting AVX2 and FMA

newpavlov:
We must do it according to the [Intel manual](https://www.intel.com/content/dam/develop/external/us/en/documents/36945). Our non-conformance has [caused](https://github.com/RustCrypto/hashes/pull/386) issues with the `sha2` crate.

Strictly speaking, during detection of AVX we also should use XGETBV to check whether XMM and YMM states are enabled by OS, but I think it's worth to do it later in a separate PR.

cc @brocaar

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -33,7 +33,7 @@ version = "0.0.2"
 
 [[package]]
 name = "cpufeatures"
-version = "0.2.2"
+version = "0.2.3"
 dependencies = [
  "libc",
 ]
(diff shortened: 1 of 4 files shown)


tarcieri (approved):

brocaar:
After merging, could you create a new (minor) release of this crate?

newpavlov:
@brocaar 
Done. Can you verify that the new version of `cpufeatures` fixes the `sha2` issue?

brocaar:
I'm currently recompiling with cpufeatures updated to v0.2.3, I will let you know if this solves the issue.

brocaar:
@newpavlov I can confirm that this update solved https://github.com/RustCrypto/hashes/pull/386.

Thanks for your support!

PR #799, opened 2022-09-03
block-buffer: add try_new method

rpiasetskyi:
try_new creates BlockBuffer or returns an error.

Requested in https://github.com/RustCrypto/traits/pull/1078

### DIFF
--- a/block-buffer/src/lib.rs
+++ b/block-buffer/src/lib.rs
@@ -8,7 +8,7 @@
 
 pub use generic_array;
 
-use core::{marker::PhantomData, slice};
+use core::{fmt, marker::PhantomData, slice};
 use generic_array::{
     typenum::{IsLess, Le, NonZero, U256},
     ArrayLength, GenericArray,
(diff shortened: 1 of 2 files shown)


block-buffer/src/lib.rs
```diff
@@ -40,6 +40,19 @@ pub type EagerBuffer<B> = BlockBuffer<B, Eager>;
...
+pub struct Error;
+
+impl fmt::Display for Error {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
+        f.write_str("Block buffer error")
+    }
+}
+
+#[cfg(feature = "std")]
+impl std::error::Error for Error {}
```
newpavlov, line 54: `block-buffer` does not have `std` feature. Since the `try_new` method is intended to be used only by our crates with immediate remap into a different error, I think we can omit implementation of the `Error` trait.

block-buffer/tests/mod.rs
```diff
@@ -186,3 +186,11 @@ fn test_eager_paddings() {
         [0x42, 0xff, 0x10, 0x11],
     );
 }
+
+#[test]
+fn test_try_new() {
+    assert!(EagerBuffer::<U4>::try_new(&[0x01, 0x02, 0x03]).is_ok());
+    assert!(EagerBuffer::<U4>::try_new(&[0x01, 0x02, 0x03, 0x04]).is_err());
+    assert!(LazyBuffer::<U4>::try_new(&[0x01, 0x02, 0x03, 0x04]).is_ok());
+    assert!(LazyBuffer::<U4>::try_new(&[0x01, 0x02, 0x03, 0x04, 0x05]).is_err());
```
newpavlov, line 195: I think `&[0; N]` would be a bit more clear.

Issue #814, opened 2022-10-20
hex-literal: v1.0 tracker


newpavlov:
`hex-literal` is quite stable, with v0.3.0 being release over two years ago. So I think we can start to consider v1.0 release.

Potential changes:
- [x] Bump MSRV to 1.56 and edition to 2021 (Done in v0.4)
- [x] Decide whether we want to forbid comments inside literals or not, as described [here](https://github.com/RustCrypto/utils/issues/813#issuecomment-1286189440) (Done in v0.4)
- [x] Use https://github.com/rust-lang/cargo/issues/9930. It will allow us to bump `hex-literal` edition without breaking downstream builds and maybe incorporate new language features.
- [x] Use inline `const {}` expressions to force const evaluation (MSRV 1.79).
- [ ] Wait for stabilization of `generic_const_exprs` and `adt_const_params`? It would allow us to write:
```rust
fn decode<const DATA: &'static [u8]>() -> [u8; decode_len(DATA)] { ... }
```

Nugine:
It's possible to implement `hex!` macro with const evaluation. See [const_str::hex!](https://docs.rs/const-str/latest/const_str/macro.hex.html).

newpavlov:
Yeah, it [works](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=4fd209f78f8868d7fcf16f895617f962) on Nightly and the code does not look too bad.

But this approach has the following disadvantages:
1) `hex::<"abcdef">()` is... not great. Introducing a wrapper macro partially solves the issue, but ideally we would use [const arguments](https://internals.rust-lang.org/t/pre-rfc-const-function-arguments/6709). Unfortunately, it's highly unlikely that they will be added to the language. :(
2) Error diagnostics are much worse. We can not even panic with message which includes encountered invalid character. With macros we potentially can point the exact place which triggers an error.

newpavlov:
hex-literal v1.0.0 is released!

PR #816, opened 2022-10-26
hex-literal: remove support for comments inside literals and migrate to CTFE

newpavlov:
Potential breaking change for v1.0, see  #814.


newpavlov:
As [suggested](https://github.com/RustCrypto/utils/issues/814#issuecomment-1438272397) by @Nugine, I've migrated this crate from proc macro to CTFE. Initially I thought it would require Nightly, but looking closer at the `const_str` crate, it can work on stable as well. It result in a lighter crate, which likely should compile significantly faster.

newpavlov:
@tarcieri 
I think we can release these changes as v0.4. WDYT?

tarcieri:
@newpavlov sure!

tarcieri (approved):

triamazikamno:
Hi, this change seem to have broken the usage of hex! in the match patterns, i.e.
```
match buf {
hex!("0a0b") => {}
```
throws now:
```
error: expected pattern, found `{`
   |
60 |         hex!("0a0b") => {
   |         ^^^^^^^^^^^^^^^^
   |         |
   |         expected pattern
   |         in this macro invocation
   |         this macro call doesn't expand to a pattern
```

newpavlov:
Unfortunately, I think it's inevitable with the CTFE approach until [implementation](https://github.com/rust-lang/rust/issues/76001) of RFC 2920 will be stabilized. You can either continue to use v0.3, or introduce named constants and use them in patterns.

triamazikamno:
I see, thank you. Will stick to the v0.3 for now

tarcieri:
Another small regression (which is easy enough to work around, but was convenient before):

https://github.com/RustCrypto/SSH/actions/runs/4599840083/jobs/8125721544?pr=92#step:5:13

```
error[E0716]: temporary value dropped while borrowed
 --> ssh-encoding/tests/decode.rs:8:21
  |
8 |     let mut bytes = hex!("42").as_slice();
  |                     ^^^^^^^^^^           - temporary value is freed at the end of this statement
  |                     |
  |                     creates a temporary value which is freed while still in use
9 |     let ret = u8::decode(&mut bytes).unwrap();
  |                          ---------- borrow later used here
```

newpavlov:
Yes, it's unfortunate. Though in such cases I usually use shorter `let mut bytes = &hex!("...");` with hinted type, if necessary.

hex-literal/src/lib.rs
```diff
@@ -1,169 +1,86 @@
...
-            unexpected => panic!("expected string literal, got `{}`", unexpected),
-        };
-        out_ts.extend(iter);
-    }
-    TokenStream::from(TokenTree::Group(Group::new(Delimiter::Bracket, out_ts)))
+#[macro_export]
+macro_rules! hex {
+    ($($s:literal)*) => {{
+        const STRINGS: &[&'static [u8]] = &[$($s.as_bytes(),)*];
+        $crate::decode::<{ $crate::len(STRINGS) }>(STRINGS)
```
Nugine, line 84: This line may not be const-promoted (or const-evaluated), which may causes performance regression.

PR #821, opened 2022-12-12
cpufeatures: support freestanding/UEFI `x86` targets

rvolosatovs:
This fixes compilation of dependent crates for these targets.

For example, `sha2`:

Before:
```
$ cargo build --target x86_64-unknown-none --no-default-features
   Compiling sha2 v0.10.6 (/home/rvolosatovs/src/github.com/rustcrypto/hashes/sha2)
LLVM ERROR: Do not know how to split the result of this operator!

error: could not compile `sha2`
$ cargo build --target x86_64-unknown-uefi --no-default-features
   Compiling sha2 v0.10.6 (/home/rvolosatovs/src/github.com/rustcrypto/hashes/sha2)
LLVM ERROR: Do not know how to split the result of this operator!

error: could not compile `sha2`
```

After:
```
$ cargo build --target x86_64-unknown-none --no-default-features
   Compiling cpufeatures v0.2.5 (/home/rvolosatovs/src/github.com/rustcrypto/utils/cpufeatures)
   Compiling sha2 v0.10.6 (/home/rvolosatovs/src/github.com/rustcrypto/hashes/sha2)
    Finished dev [optimized + debuginfo] target(s) in 0.19s
$ cargo build --target x86_64-unknown-uefi --no-default-features
   Compiling cpufeatures v0.2.5 (/home/rvolosatovs/src/github.com/rustcrypto/utils/cpufeatures)
   Compiling sha2 v0.10.6 (/home/rvolosatovs/src/github.com/rustcrypto/hashes/sha2)
    Finished dev [optimized + debuginfo] target(s) in 0.19s
```

### DIFF
--- a/cpufeatures/src/x86.rs
+++ b/cpufeatures/src/x86.rs
@@ -14,11 +14,11 @@ macro_rules! __unless_target_features {
     ($($tf:tt),+ => $body:expr ) => {{
         #[cfg(not(all($(target_feature=$tf,)*)))]
         {
-            #[cfg(not(target_env = "sgx"))]
+            #[cfg(not(any(target_env = "sgx", target_os = "none", target_os = "uefi")))]
             $body
 
-            // CPUID is not available on SGX targets
-            #[cfg(target_env = "sgx")]
+            // CPUID is not available on SGX, freestanding and UEFI targets
+            #[cfg(any(target_env = "sgx", target_os = "none", target_os = "uefi"))]
             false
         }
 


tarcieri (approved):

newpavlov:
I am not sure it's a correct approach. The problem is not with the CPUID instruction. Both `none` and `uefi` targets handle it just [fine](https://rust.godbolt.org/z/7dxnsnnsz). For reference, Rust's `std` also [assumes](https://doc.rust-lang.org/src/core/up/up/stdarch/crates/core_arch/src/x86/cpuid.rs.html#101-163) that CPUID is available on all non-SGX x86-64 targets.

The issue is [caused](https://rust.godbolt.org/z/fnMd9eqYr) by the SHA instructions. To me it looks like an LLVM issue. Have you tried to look into it?

For now, you can use the [`force-soft`](https://github.com/RustCrypto/hashes/blob/master/sha2/Cargo.toml#L35) feature as a temporary workaround.

tarcieri (changes requested):
Oh yeah, interesting... for some reason I thought UEFI didn't have access to CPUID but that seems to be wrong and there are ample resources that describe calling CPUID from UEFI

rvolosatovs:
Thanks, I'll look into LLVM issue

tarcieri:
It seems people are still hitting this, even with the `force-soft` feature of `sha2` enabled, and even with the force soft cfg attributes for all of the other crates enabled:

https://github.com/RustCrypto/SSH/issues/91

I think we should really try to get a workaround for this added. We've had dozens of reports and it causes very confusing errors.

yaoxin1995:
I want to use this crate in `no_std` environment. 

However, I got the error `LLVM ERROR: Do not know how to split the result of this operator!`. 

How I add this crate:
```
ssh-key  = {version = "0.5.1", default-features = false, features = ["alloc", "ecdsa", "ed25519", "p384"]}
```

I already add `sha2` with feature `force-soft` enabled:
```
sha2 = { version = "0.10.6", default-features = false,  features = ["force-soft"]}
```

Target information in file `x86_64-qkernel.json`:
```
{
  "llvm-target": "x86_64-unknown-none",
  "data-layout": "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-f80:128-n8:16:32:64-S128",
  "linker-flavor": "gcc",
  "target-endian": "little",
  "target-pointer-width": "64",
  "target-c-int-width": "32",
  "arch": "x86_64",
  "os": "none",
  "disable-redzone": true,
  "features": "-mmx,-avx,-sse,+soft-float",
  "frame-pointer": "always"
}
```

How I compile the project:
```
RUSTFLAGS="--cfg aes_force_soft --cfg polyval_force_soft --cfg log" CARGO_TARGET_DIR=../target cargo xbuild --target x86_64-qkernel.json --release --verbose
```

Do you know how to solve it?

tarcieri:
@yaoxin1995 in theory this PR will solve it. Perhaps you could test it out?

Add the following to your Cargo.toml:

```toml
[patch.crates-io.cpufeatures]
git = "https://github.com/rvolosatovs/utils.git"
branch = "freestanding-uefi"
```

cpufeatures/src/x86.rs
```diff
@@ -14,11 +14,11 @@ macro_rules! __unless_target_features {
     ($($tf:tt),+ => $body:expr ) => {{
         #[cfg(not(all($(target_feature=$tf,)*)))]
         {
-            #[cfg(not(target_env = "sgx"))]
+            #[cfg(not(any(target_env = "sgx", target_os = "none", target_os = "uefi")))]
             $body
 
-            // CPUID is not available on SGX targets
-            #[cfg(target_env = "sgx")]
+            // CPUID is not available on SGX, freestanding and UEFI targets
```
newpavlov, line 20: This comment is a bit misleading. If I am not mistaken, CPUID is available on the later two targets, the issue is with compilation flags used by default on those targets which disable ability to use SIMD instructions. I wrote more about it here: https://internals.rust-lang.org/t/18169

newpavlov (approved):

cpufeatures/src/x86.rs
tarcieri, line 20: How about this?

```suggestion
            // CPUID is not available on SGX. Freestanding and UEFI targets
            // do not support SIMD features with default compilation flags.
```

newpavlov: Sounds good to me.

tarcieri (approved):

PR #831, opened 2023-01-23
CryptoPAN 0.1.0-pre

mikemcdermottredjack:
Anonymizes IP addresses using the CryptoPAN algorithm tightly based on the GO implementation by Yawning Angel (https://github.com/Yawning/cryptopan), which is based on the original reference implementation [paper by J. Fan, J. Xu, M. Ammar, and S. Moon. (https://ieeexplore.ieee.org/abstract/document/1181415)]

CryptoPAN is a prefix-preserving, 1-1 mapping algorithm that allows for consistent anonymization of IP addresses across datasets, provided that the same 256-bit key is used. 

IPv6 anonymization is supported, but it is not known if the code conforms to the reference implementation.


tarcieri:
This might be a better fit for the https://github.com/rustcrypto/formats repo /cc @newpavlov

cryptopan/CHANGELOG.md
```diff
@@ -0,0 +1,8 @@
+<h1>Changelog</h1>
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to [Semantic Versioning]https://semver.org/spec/v2.0.0.html
+
+<h1>0.1.0-PRE (2022-01-23)</h1>
+<h2>Added</h2>
```
newpavlov, line 7: You should use `#` and `##` instead of the HTML tags, see the other crate for reference.

cryptopan/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+categories = ["cryptography"]
+edition = "2021"
+readme = "README.md"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+aes="0.8.1"
+bitvec = "1.0.1"
+ecb="0.1.1"
```
newpavlov, line 18: You don't need the `ecb` crate anymore. `BlockEncrypt`/`BlockDecrypt` trait impls on cipher types can be used to perform ECB. Also please add trailing newline.

cryptopan/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+authors = ["RustCrypto Developers"]
+license = "MIT OR Apache-2.0"
+description = "CryptoPAN algorithm for IP address anonymization"
+repository = "https://github.com/RustCrypto/utils"
+keywords = ["cryptopan", "ip anonymization"]
+categories = ["cryptography"]
+edition = "2021"
+readme = "README.md"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
newpavlov, line 13: Remove this comment.

cryptopan/README.md
```diff
@@ -0,0 +1,8 @@
+<h1>CryptoPAN IP Address Anonymization</h1>
```
newpavlov, line 1: Use `#` instead of `<h1>`. Also please add badges and other sections as in other crate READMEs.

cryptopan/src/cryptopan.rs
```diff
@@ -0,0 +1,225 @@
...
+use bitvec::{bitvec, vec::BitVec};
+use std::net::{
+    IpAddr,
+    IpAddr::{V4, V6},
+    Ipv4Addr, Ipv6Addr,
+};
+
+type Aes128EcbEnc = ecb::Encryptor<aes::Aes128>;
+
+pub struct CryptoPAn {
```
newpavlov, line 21: Maybe make this type generic over used ciphers with AES-128 being the default one? Also add some docs to the type.

cryptopan/src/cryptopan.rs
```diff
@@ -0,0 +1,225 @@
...
+    aes: Aes128EcbEnc,
+    pad: BitVec<u8, Msb0>,
+}
+
+impl CryptoPAn {
+    /// Creates new outer block for encryption using key. Panics if key size isn't 256 bits
+    pub fn from_key(key: &[u8]) -> Self {
```
newpavlov, line 28: I think this method should accept `&[u8; 32]`. Alternatively, it could implement the [`KeyInit`](https://docs.rs/cipher/0.4.3/cipher/trait.KeyInit.html) trait.

cryptopan/src/cryptopan.rs
```diff
@@ -0,0 +1,225 @@
...
+        // Split AES key into two parts - one is padding
+        // Encrypt the padding with the first part of the key
+        let mut aes = Aes128EcbEnc::new(GenericArray::from_slice(&key[0..16]));
+
+        let mut buf = [0u8; 16];
+        buf.copy_from_slice(&key[16..32]);
+        let g = GenericArray::from_mut_slice(&mut buf);
```
newpavlov, line 35: You could use `GenericArray::clone_from_slice(&key[16..32])` here.

cryptopan/src/cryptopan.rs
```diff
@@ -0,0 +1,225 @@
...
+        aes.encrypt_block_mut(g);
+
+        Self {
+            aes,
+            pad: BitVec::from_slice(g.as_slice()),
+        }
+    }
+    /// Creates new outer block for encryption using str key. Panics if key size isn't 256 bits.
+    pub fn from_str_key(string_key: &str) -> Self {
```
newpavlov, line 44: This method looks redundant to me, since I think it will be better in user code that a string is used as raw bytes.

cryptopan/src/cryptopan.rs
```diff
@@ -0,0 +1,225 @@
...
+            76u8, 45u8, 42u8, 132u8, 34u8, 2u8,
+        ])
+    }
+
+    #[test]
+    fn test_encrypt_ipv4() {
+        // This is a copy of the tests for the Go implementation
+        // https://github.com/Yawning/cryptopan/blob/master/cryptopan_test.go
+        // IPV4 tests
+        let ipv4 = HashMap::from([
```
newpavlov, line 119: You do not need `HashMap` here. You can keep the tests in the array form.

cryptopan/src/cryptopan.rs
```diff
@@ -0,0 +1,225 @@
...
+                get_key()
+                    .encrypt(V4(Ipv4Addr::from_str(k).unwrap()))
+                    .to_string()
+            );
+        }
+    }
+
+    #[test]
+    fn test_encrypt_ipv6() {
+        let ipv6 = HashMap::from([
```
newpavlov, line 204: Same here.

cryptopan/src/cryptopan.rs
```diff
@@ -0,0 +1,225 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::{collections::HashMap, str::FromStr};
+
+    fn get_key() -> CryptoPAn {
+        CryptoPAn::from_key(&[
+            21u8, 34u8, 23u8, 141u8, 51u8, 164u8, 207u8, 128u8, 19u8, 10u8, 91u8, 22u8, 73u8,
+            144u8, 125u8, 16u8, 216u8, 152u8, 143u8, 131u8, 121u8, 121u8, 101u8, 39u8, 98u8, 87u8,
```
newpavlov, line 109: You don't need to repeat `u8` for every element. Also I think it will be a bit more idiomatic to define a constant:
```rust
const KEY: [u8; 32] = [21, 34, ..];
```
You also could use the [`hex-literal`](https://docs.rs/hex-literal) crate, but I guess it's a bit redundant to bring it for just one constant.

cryptopan/src/lib.rs
```diff
@@ -0,0 +1 @@
+pub mod cryptopan;
```
newpavlov, line 1: Simply define all the code in `lib.rs` without splitting it into a separate module. Also note that module level docs (i.e. the `//!` will be used as documentation).

cryptopan/src/cryptopan.rs
```diff
@@ -0,0 +1,225 @@
...
+                bs[80..96].load::<u16>(),
+                bs[96..112].load::<u16>(),
+                bs[112..128].load::<u16>(),
+            ))
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
newpavlov, line 102: I think it would be better to move these integration tests into the `tests/` folder.

cryptopan/src/cryptopan.rs
```diff
@@ -0,0 +1,225 @@
...
+            aes,
+            pad: BitVec::from_slice(g.as_slice()),
+        }
+    }
+    /// Creates new outer block for encryption using str key. Panics if key size isn't 256 bits.
+    pub fn from_str_key(string_key: &str) -> Self {
+        Self::from_key(string_key.as_bytes())
+    }
+    /// Encrypts either a V4 or V6 IP address using the key
+    pub fn encrypt(&self, input: IpAddr) -> IpAddr {
```
newpavlov, line 48: I think it will be better to define separate `encrypt_v4` and `encrypt_v6` methods, making `encrypt` a simple matching wrapper around them. Also I think we can do without `bitvec`.

newpavlov (commented):
Hm, I am not sure about into which repository it should go, but I am fine with the formats repo as well.

cryptopan/src/cryptopan.rs
mikemcdermottredjack: Would you just shift bits to do the operations instead of using `bitvec`? I think using `bitvec` is cleaner and easier to understand, but I understand replacing it if it performs more poorly than working with the bits directly. Or perhaps there is another reason not to use it?

cryptopan/README.md
mikemcdermottredjack: I added the badges except for crates.io, docs.rs, and build automation (since it has not been published to any of those systems)

tarcieri:
@mikemcdermottredjack before we get too deep into code review, can you reopen this PR under https://github.com/rustcrypto/formats ?

mikemcdermottredjack:
@tarcieri sure

PR #832, opened 2023-01-26
adding a zeroize trait for blockbuffers so that blake2 can have zeroize too (not zeroizeondrop- just zeroize)

laudiacay:


laudiacay:
oof. what should i do about this namespaced feature/nightly thing? @tarcieri

laudiacay:
https://github.com/RustCrypto/hashes/issues/87#issuecomment-513800633

the blockbuffer doesn't expose the underlying genericarray, or even allow you to have a mutable slice to hook zeroize up to. there are some weird setter methods but i am concerned that might get optimized out by the compiler. 

i could alternatively add a method to pull a mutable slice out of mutable blockbuffers, or expose a genericarray, let me know which one is preferable to this.

block-buffer/Cargo.toml
```diff
@@ -14,6 +14,10 @@ readme = "README.md"
 [dependencies]
 crypto-common = "0.2.0-pre"
 generic-array = "0.14"
+zeroize = { version = "1.4", optional = true, features = ["zeroize_derive"] }
 
 [dev-dependencies]
 hex-literal = "0.3.3"
+
+[features]
+zeroize = ["dep:zeroize"]
```
tarcieri, line 23: If you just remove the entire `[features]` section I'd think it would be fine.

Namespaced features aren't MSRV-compatible, as they require Rust 1.60. The MSRV being tested in CI is 1.56 (although it doesn't seem to be correct in the README.md)

Issue #743, opened 2022-03-01
Zeroize performance on u8 arrays


sopium:
I inspected the generated assembly code and benchmarked `zeroize` for `[u8; 32]` on x86_64 and found it quite inefficient, storing one byte at a time:

https://play.rust-lang.org/?version=stable&mode=release&edition=2021&gist=3f44f4b90e6af0eac0dcb1f649390329

On my Ryzen CPU, it takes ~7.8324 ns, or ~1cpb. Binary code size is also quite large.

Using inline assembly (just stabilized in 1.59) and SSE2, zeroing a `[u8; 32]` takes just 3 instructions and ~492.87 ps (~16 bytes per cycle):

```rust
let mut buf: [u8; 32];
core::arch::asm!(
    "xorps {zero}, {zero}",
    "movups {zero}, ({ptr})",
    "movups {zero}, 16({ptr})",
    zero = out(xmm_reg) _,
    ptr = in(reg) &mut buf,
    options(att_syntax, nostack, preserves_flags),
);
```

So it might be something worth optimizing/documenting.

If you do not want to use inline assembly, maybe you should encourage using larger types or SIMD types, e.g., `[u64; 4]` or `[__m128; 2]` instead of `[u8; 32]`. Using `write_volatile` on `*mut __m128` generates equally compact and efficient code as the assembly code above.

tarcieri:
@sopium we've done practically no optimization work, so I'm glad you're thinking about it.

Inline assembly is fine, but would need to be feature gated due to the higher MSRV (we can eventually consider bumping it, but our last bump was already painful)

For something like `[T; N]` though, it would be good to ensure this doesn't monomorphize to `N` distinct blobs of ASM.

sopium:
It boils down to LLVM not able to optimize `write_volatile` for u8 pointers. The bytes have to be written individually.

I think monomorphize for 16/32/64-byte arrays should be fine. These would also be commonly used key sizes in cryptography so the optimization can be justified.

For a u8 slice with unknown but largish size you can also use `align_to_mut`:

```rust
pub fn zeroize_slice(x: &mut [u8]) {
    unsafe {
        // Use __m128 on x86/x86_64.
        let (p, m, s) = x.align_to_mut::<u64>();
        for b in p {
            core::ptr::write_volatile(b, 0);
        }
        for m in m {
            core::ptr::write_volatile(m, 0);
        }
        for b in s {
            core::ptr::write_volatile(b, 0);
        }
    }
}
```

[...]

PR #841, opened 2023-02-28
improve performance for slice zeroization (issue #743)

cbeck88:
The purpose of the change is to make calls to `x.as_mut_slice().zeroize()` considerably faster, particularly for types like `[u8; n]`. We take @sopium's proposed code from #743 without significant changes.

The reason it becomes faster is that the call to `volatile_set` before this change appears not to be easily optimizable, and (for example) leads to setting bytes one at a time, instead of the compiler consolidating them into SIMD instructions.

In the modified code, we don't use `volatile_set`, we instead loop over the slice setting the elements to `Default::default()`, and to ensure that the writes are not optimized out, we use an empty asm block. (There is discussion of the correct asm options to use here in the issue.)

Because the asm block potentially reads from the pointer and makes a syscall of some kind, the compiler cannot optimize out the zeroizing, or it could cause observable side-effects. In the improved code, we only create such an optimization barrier once, rather than after each byte that it is written.

The call to `atomic_fence()` is not changed.

---

This change may help give users a way to improve performance, if they have to zeroize very large objects, or, frequently have to zeroize many small objects. We tested code-gen here in godbolt (in addition to the tests posted in the github issue) and found that this change is typically enough for llvm to start adding in SIMD optimizations that zero many bytes at once.

### DIFF
--- a/zeroize/src/lib.rs
+++ b/zeroize/src/lib.rs
@@ -471,7 +471,17 @@ where
         // object for at least `self.len()` elements of type `Z`.
         // `self.len()` is also not larger than an `isize`, because of the assertion above.
         // The memory of the slice should not wrap around the address space.
-        unsafe { volatile_set(self.as_mut_ptr(), Z::default(), self.len()) };
+        for z in self.iter_mut() {
+            *z = Z::default();
+        }
+        unsafe {
+            core::arch::asm!(
+                "/* {ptr} */",
+                ptr = in(reg) self.as_mut_ptr(),
+                options(nostack, readonly, preserves_flags),
+            );
+        }
+
         atomic_fence();
     }
 }


tarcieri:
This should probably be feature gated to avoid a massive MSRV bump and disrupting existing users. That was very problematic the last time we bumped MSRV to add const generic support.

And if all the inline ASM is doing is providing an optimization barrier, it seems like `core::hint::black_box` could be used instead.

cbeck88:
I'm a little worried about this part of the documentation for `core::hint::black_box`: https://doc.rust-lang.org/beta/core/hint/fn.black_box.html#when-is-this-useful

tarcieri:
Oh wow, definitely an important detail about `core::hint::black_box`!

An ASM optimization barrier seems good then, although let me run this implementation by a few people.

It would definitely still be good to feature gate it in order to preserve MSRV.

tarcieri:
Hmm, when did that warning get added?

It doesn't appear on the current stable docs. Is it new?

https://doc.rust-lang.org/stable/std/hint/fn.black_box.html

cbeck88:
Maybe I looked up the wrong docs, my url says "beta". Maybe they removed that later.

cbeck88:
Maybe we can look up the implementation, if it's very similar to Sopium's suggested barrier then maybe it's fine

tarcieri:
> Maybe I looked up the wrong docs, my url says "beta". Maybe they removed that later.

To me that implies those docs were recently added. I guess we'll see what happens in the next release.

zeroize/src/lib.rs
```diff
@@ -471,7 +471,17 @@ where
...
-        unsafe { volatile_set(self.as_mut_ptr(), Z::default(), self.len()) };
+        for z in self.iter_mut() {
+            *z = Z::default();
+        }
+        unsafe {
+            core::arch::asm!(
+                "/* {ptr} */",
+                ptr = in(reg) self.as_mut_ptr(),
+                options(nostack, readonly, preserves_flags),
+            );
```
tarcieri, line 482: `asm!` is only stable for x86/x86-64,  ARM/AArch64, and RISC-V, so its usage needs to be gated for those platforms

elichai:
Worth following some discussion here: https://rust-lang.zulipchat.com/#narrow/stream/122651-general/topic/black_box.20and.20crypto

zeroize/src/lib.rs
MasterAwesome: https://github.com/RustCrypto/utils/blob/f8f6f6e703a62dcbe0cc996c9b77bac2a9e9ad9f/zeroize/README.md?plain=1#L32

This guarantee needs to be updated if this change is merged.

tarcieri: Elsewhere we feature-gate `asm`, so we could perhaps maintain that guarantee so long as the `asm` feature is off

newpavlov:
This PR is quite stale, so I think we can close it? Additionally, I think #1269 would be a better direction for resolving the performance issues.

tarcieri:
As has been discussed in #1252 (including being pointed out by @RalfJung) as well as #1269, using `asm!` to perform the writes provides actual guarantees that they will not be removed (and can provide better performance than `volatile_write`), whereas the `optimization_barrier` approach does not.

That said, I agree this PR is stale and not the correct approach.

PR #842, opened 2023-03-02
zeroize: impl Zeroize for `str` and `Box<str>`

Nugine:


zeroize/src/lib.rs
```diff
@@ -476,6 +476,14 @@ where
     }
 }
 
+impl Zeroize for str {
+    fn zeroize(&mut self) {
+        // Safety:
+        // A zeroized byte slice is a valid UTF-8 string.
+        unsafe { self.as_bytes_mut().zeroize() }
```
tarcieri, line 483: I guess this is the best you can do

PR #848, opened 2023-03-16
cpufeatures: Support dynamic feature detection on iOS and derivative platforms

complexspaces:
This PR adds support for dynamic feature detection on iOS, watchOS, and tvOS, bringing full support to all of Apple's platforms and following up https://github.com/RustCrypto/utils/pull/435.

I'm not entirely sure what the previous comment about lack of `sysctl(3)` support meant, FWIW. iOS supports the same kernel calls that macOS does (for the most part, but especially low-level ones like this) because they're based on the same kernel and, more importantly, the function appears in the public headers inside each platform's SDK.

To confirm/prove that there wasn't any problematic filtering occurring on iOS, I ran these changes on both the iOS simulator and on a real device (iPad Pro 3rd gen). The results make sense AFAICT because this specific model runs a full-fledged M1 processor in it.

Test code:
```rust
cpufeatures::new!(armcaps, "aes", "sha2", "sha3");

let feature_set = armcaps::init();
if feature_set.get() {
    log::info!("we had support for AES, SHA2, and SHA3 instructions")
} else {
   log::info!("no support for SHA3");
}
```

Results in iOS simulator:
```
INFO  2023-03-16T22:20:04.778 ThreadId(1) we had support for AES, SHA2, and SHA3 instructions
```

Results on physical device:
```
INFO  2023-03-16T22:33:11.875 ThreadId(1) we had support for AES, SHA2, and SHA3 instructions
```

### DIFF
--- a/cpufeatures/Cargo.toml
+++ b/cpufeatures/Cargo.toml
@@ -14,7 +14,7 @@ categories = ["no-std"]
 edition = "2018"
 readme = "README.md"
 
-[target.aarch64-apple-darwin.dependencies]
+[target.'cfg(all(target_arch = "aarch64", target_vendor = "apple"))'.dependencies]
 libc = "0.2.95"
 
 [target.'cfg(all(target_arch = "aarch64", target_os = "linux"))'.dependencies]
--- a/cpufeatures/src/aarch64.rs
+++ b/cpufeatures/src/aarch64.rs
@@ -37,8 +37,8 @@ pub fn getauxval_hwcap() -> u64 {
     unsafe { libc::getauxval(libc::AT_HWCAP) }
 }
 
-// MacOS runtime detection of target CPU features using `sysctlbyname`.
-#[cfg(target_os = "macos")]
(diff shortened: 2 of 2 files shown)


tarcieri:
#378 was the previous issue discussing this

complexspaces:
I read through that issue, but I didn't see any specific mention of iOS unavailability (nor the PR I linked beyond the same comment that existed in the code). Did I miss something in that issue?

complexspaces:
> Is there any way of adding CI support?

That's a good question. I don't know if there's actually a way to test this implementation today with GitHub Actions as they don't yet offer hosted M1 runners (Rust is [also waiting on this](https://github.com/rust-lang/rust/issues/73908)). Without those, there's no way to run the `aarch64-apple-ios-sim` simulator via XCode.

> If it works that's great, but I have no way of confirming.

Is this something you would consider a blocker for merging? I can attempt to collect a more live demo if it would be considered useful.

tarcieri:
I can go ahead and merge this and we can see if complaints pop up or not.

tarcieri (approved):

PR #858, opened 2023-03-23
Update to syn-2

maurer:
This won't test cleanly or go in yet, as it depends on synstructure updating to support syn-2 (this is [trivial](https://github.com/mystor/synstructure/pull/57), but hasn't landed yet).

Uploading the code for review in case you have feedback before it's landable.

### DIFF
--- a/fiat-constify/src/main.rs
+++ b/fiat-constify/src/main.rs
@@ -5,16 +5,16 @@
 
 #![allow(clippy::single_match, clippy::new_without_default)]
 
-use proc_macro2::Span;
+use proc_macro2::{Literal, Span};
 use quote::{quote, ToTokens};
 use std::{collections::BTreeMap as Map, env, fs, ops::Deref};
 use syn::{
     punctuated::Punctuated,
-    token::{Bang, Brace, Bracket, Colon, Const, Eq, Let, Mut, Paren, Pound, RArrow, Semi},
+    token::{Brace, Bracket, Colon, Const, Eq, Let, Mut, Not, Paren, Pound, RArrow, Semi},
     AttrStyle, Attribute, Block, Expr, ExprAssign, ExprCall, ExprLit, ExprPath, ExprReference,
-    ExprRepeat, ExprTuple, FnArg, Ident, Item, ItemFn, ItemType, Lit, LitInt, Local, Pat, PatIdent,
-    PatTuple, PatType, Path, PathArguments, PathSegment, ReturnType, Stmt, Type, TypeArray,
-    TypePath, TypeReference, TypeTuple, UnOp,
+    ExprRepeat, ExprTuple, FnArg, Ident, Item, ItemFn, ItemType, Lit, LitInt, Local, LocalInit,
+    MacroDelimiter, Meta, MetaList, Pat, PatIdent, PatTuple, PatType, Path, PathArguments,
+    PathSegment, ReturnType, Stmt, Type, TypeArray, TypePath, TypeReference, TypeTuple, UnOp,
 };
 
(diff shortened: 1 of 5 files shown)


maurer:
I missed the edit and wrote a commit removing the synstructure dep. Do you want me to drop that commit or leave it in?

tarcieri:
Oh, if you’ve already done it one fewer dependency is great, thank you!

tarcieri:
Hmm, seems like the `zeroize` tests are failing on fields that use `#[zeroize(skip)]` because they aren't being accessed.

I think you could go ahead and add `#[allow(unused_variables)]` to those fields?

maurer:
Added the #[allow(unused_variables)], and put a `#![deny(warnings)]` into the zeroize derive test that was breaking - it had scrolled past clean locally because the warnings got pushed off the screen, but you must be overriding that in CI somewhere. That should make it fail locally for others.

tarcieri:
We deny warnings in CI but allow them locally to make refactoring easier. I'd suggest removing the `#![deny(warnings)]`.

tarcieri:
Hmm, seems this will entail an MSRV bump for the derive feature. That's fine.

tarcieri:
I opened #869 which does a 2021 edition upgrade for `zeroize` and `zeroize_derive`

tarcieri:
@maurer if you rebase it should take care of the MSRV-related issues

maurer:
Done.

zeroize/derive/src/lib.rs
```diff
@@ -253,277 +280,278 @@ impl ZeroizeAttrs {
     }
 }
 
-fn generate_fields(s: &mut synstructure::Structure<'_>, method: TokenStream) -> TokenStream {
-    s.bind_with(|_| BindStyle::RefMut);
+fn field_ident(n: usize, field: &Field) -> Ident {
+    if let Some(ref name) = field.ident {
+        name.clone()
+    } else {
+        format_ident!("__zeroize_field_{n}")
```
tarcieri, line 287: That seems odd... perhaps this?

```suggestion
        format_ident!("__zeroize_field_{}", n)
```

fiat-constify/src/main.rs
```diff
@@ -59,29 +59,33 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
 
 /// Build a toplevel attribute with the given name and comma-separated values.
 fn build_attribute(name: &str, values: &[&str]) -> Attribute {
-    let span = Span::call_site();
     let values = values
         .iter()
         .map(|value| build_path(value))
         .collect::<Vec<_>>();
+    let path = build_path(name);
+    let tokens = quote! { (#(#values),*) };
```
tarcieri, line 67: I spot checked the generated code and this appears to be adding an extra level of paren nesting.

Here's a diff of the generated code:

```
```
-#![allow(
-    dead_code,
-    rustdoc::bare_urls,
-    rustdoc::broken_intra_doc_links,
-    unused_assignments,
-    unused_mut,
-    unused_variables
-)]
+# ! [allow ((clippy :: identity_op , clippy :: unnecessary_cast , dead_code , rustdoc :: broken_intra_doc_links , unused_assignments , unused_mut , unused_variables))]
```
```

tarcieri, line 67: I confirmed locally this works:

```suggestion
    let tokens = quote! { #(#values),* };
```

maurer: You're right, the data structure changed between the two versions and I missed it, thanks.

tarcieri:
Awesome, thanks!

PR #861, opened 2023-03-24
cmov, fiat-constify, hybrid-array, zeroize: fix CI targets and branches

brxken128:
Some `--target` arguments were not being provided to `cargo test`, which meant CI was falsely passing for the `cmov` crate (`x86` currently does not build, I'll open another PR shortly to fix that).

This also updates some `branches: main` references to `branches: master`, and adds the `cmov` workflow path into the workflow itself.


tarcieri:
@brxken128 can you rebase now that #863 has landed?

PR #863, opened 2023-03-24
cmov: fix builds on `x86` (32-bit) targets

brxken128:
There was a bug with CI (#861) which meant build issues on any architecture that differs from the host (`x86_64` I believe) were suppressed.

I think this specific bug happens due to `x86` registers not having the capacity to store anything > `i32`. To fix it, I re-used the `cmov` implementation for `u128`, but re-purposed it specifically for `x86` and `u64`s.


cmov/src/x86.rs
```diff
@@ -49,6 +49,7 @@ impl Cmov for u32 {
     }
 }
 
+#[cfg(not(target_arch = "x86"))]
```
tarcieri, line 52: I think this would be more clearly expressed as:

```suggestion
#[cfg(target_arch = "x86_64")]
```

brxken128: I agree, it's easier to understand for sure (especially because there's only two possible options within this module).

PR #870, opened 2023-03-25
Add RawPadding trait to block-padding for padding blocks of arbitrary size

### LINKED ISSUE #860: block-padding does not support dynamic block sizes
It used to support this, but now that it uses BlockSize/Block, it’s no longer possible to set the block size dynamically. I don’t think the added safety of requiring the block size to be specified at compile time outweigh the costs of reduced functionality.

See https://stackoverflow.com/questions/75834956/is-it-possible-to-use-a-dynamic-block-size-with-block-padding for some additional context.

czinn:
Closes #860 

The existing `Padding` trait is automatically implemented using `raw_pad` and `raw_unpad` from `RawPadding`.

Because all implementations are inlined, code generation for users of `Padding` is the same as before. I checked this by defining some functions (see below) to monomorphize the `Padding` trait for an instance of `Padding` and comparing the generated assembly (using e.g. `cargo asm block_padding::concrete_pad --build-type release`) before and after this change. I observed no differences.

``` rust
pub fn concrete_pad(block: &mut Block<generic_array::typenum::U128>, pos: usize) {
    Iso7816::pad(block, pos);
}

pub fn concrete_unpad(block: &mut Block<generic_array::typenum::U128>) -> Result<&[u8], UnpadError> {
    Iso7816::unpad(block)
}
```

I'm not attached to `RawPadding` as a name, something like `ArbitraryPadding` could also work. I think `Padding` and `BlockPadding` for (what is currently called) `RawPadding` and `Padding` would be better, but it wouldn't be worth the breaking interface change.

### DIFF
--- a/block-padding/CHANGELOG.md
+++ b/block-padding/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## 0.3.3 (2023-03-25)
+### Added
+- `RawPadding` trait for padding blocks of arbitrary size ([#870])
+
+[#870]: https://github.com/RustCrypto/utils/pull/870
+
 ## 0.3.2 (2022-03-10)
 ### Fixed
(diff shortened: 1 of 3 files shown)


newpavlov:
Thank you! I will publish this crate in the current form, but I probably will redesign the trait design in v0.4.

PR #871, opened 2023-03-25
cmov: make `Cmov::cmovz` a provided method

tarcieri:
This makes it easier to implement the `Cmov` trait for foreign types.

It additionally makes the `value` argument an `&Self` reference, which should make it easier to impl the trait for non-`Copy` types where it's desirable to avoid a move.

Note: breaking change

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -27,7 +27,7 @@ dependencies = [
 
 [[package]]
 name = "cmov"
-version = "0.2.0"
+version = "0.3.0-pre"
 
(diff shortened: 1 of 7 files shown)


brxken128:
These changes look awesome! Definitely makes it easier to implement externally so I'm all for it.

brxken128 (commented):
Everything looks good to me! The API is definitely much cleaner now, which is definitely a plus (and implementing is even easier).

Issue #872, opened 2023-03-25
cmov: implement a constant-time equality trait


brxken128:
In #868, the idea of a constant-time equality trait within `cmov` was discussed.

Rust currently has quite a _lot_ of constant-time equality traits, so I'm unsure if bringing a new one into the field is the best idea. 

It's possible that we could add support for it within `cmov`, but then implement it into a crate such as `subtle` (likely behind a new feature). This approach means that we don't bring yet another CT-EQ trait into the field, but can still provide constant-time equality comparisons that won't be optimized away by the compiler. Current users of the `subtle` crate could also enable this with almost no-effort.

I have a working `x86` PoC locally - it does not seem out of reach for `Aarch64` either.

I'd love to hear any thoughts/ideas surrounding this, as there's a lot applications where this functionality would be beneficial.

tarcieri:
Something like this possibly?

```rust
pub trait CmovEq {
    fn cmoveq(a: &Self, b: &Self, condition: &mut Condition);
}
```

brxken128:
I think that'd be pretty suitable - it's in-line with the current API, and low-level enough to be applicable to pretty much any application that I can think of.

Also would be quite easy to make use of it externally.

brxken128:
As a bit of a progress update, I have it working extremely well for `x86` (currently impl'd for `u8` and `u16` but further additions should be simple).

Should we have a `cmovne` function also? We could always leave it up to the caller if needs be, but implementing it wouldn't be hard at all.

tarcieri:
Sure, though it would be good to make it a provided method

brxken128:
I've made quite a lot of progress! I have working implementations for both `x86` and `Aarch64`.

Some of the code is pretty unoptimised currently, but it works - [here's the branch](https://github.com/RustCrypto/utils/compare/master...brxken128:utils:cmoveq).

I'm a little unsure on the provided method, and on the `u128` implementation - they seem pretty clunky in comparison. Definitely still have some work to go haha

tarcieri:
I think it would probably make more sense if `CmovEq::cmovne` were provided and `CmovEq::cmoveq` were the one you implement

brxken128:
> I think it would probably make more sense if `CmovEq::cmovne` were provided and `CmovEq::cmoveq` was the one you implement

I agree - will change that. Should I open a PR so we can work on cleaning it up? Some of the implementation is a little sub-par, but I'm not too sure on how to improve it.

[...]

PR #873, opened 2023-03-26
cmov: implement constant-time equality comparisons

brxken128:
Following on from #872, this is heavily a WIP, and a lot of the implementation is sub-par.

I cleaned the macro usage up, although it seemed intentionally designed that way so I can revert those changes if needs be.

I'm not 100% sure on the portable implementation, nor am I sure that `EOR` is the most suitable instruction for Aarch64 (but it works).

Additionally, I think we should note that direct usage won't provide the most ideal outcomes - I believe the compiler will still attempt to optimise in some places, although I'm not 100% sure on that. Using this crate within `subtle` (or similar) sounds the most suitable to me.

Adding benches via `dudect_bencher` could be very handy, just as a sanity-check (but I'd be happy to hear any ideas regarding that).

Edit: in hindsight, I'm not sure that dereferencing within the macros is the best approach either (implementation-wise).

Feel free to rip this work to shreds - it's best that we get it right and make no mistakes 🙏

### DIFF
--- a/cmov/src/aarch64.rs
+++ b/cmov/src/aarch64.rs
@@ -1,4 +1,4 @@
-use crate::{Cmov, Condition};
+use crate::{Cmov, CmovEq, Condition};
 use core::arch::asm;
 
 macro_rules! csel {
@@ -14,6 +14,28 @@ macro_rules! csel {
                 options(pure, nomem, nostack),
             };
         }
+    };
+}
+
+macro_rules! csel_eq {
+    ($instruction:expr, $lhs:expr, $rhs:expr, $condition:expr, $dst:expr) => {
+        let mut tmp = *$dst as u16;
+        unsafe {
+            asm! {
+                "eor {0:w}, {1:w}, {2:w}",
+                "cmp {0:w}, 0",
+                "csel {3:w}, {4:w}, {5:w}, NE",
+                out(reg) _,
+                in(reg) *$lhs,
+                in(reg) *$rhs,
+                inlateout(reg) tmp,
+                in(reg) $condition as u16,
+                in(reg) tmp,
+                options(pure, nomem, nostack),
+            };
+        };
+
+        *$dst = tmp as u8;
     };
 }
 
@@ -41,6 +63,12 @@ impl Cmov for u16 {
     }
 }
 
+impl CmovEq for u16 {
+    fn cmovne(&self, rhs: &Self, input: Condition, output: &mut Condition) {
+        csel_eq!("csel {2:w}, {3:w}, {4:w}, NE", self, rhs, input, output);
+    }
+}
+
 impl Cmov for u32 {
     #[inline(always)]
     fn cmovnz(&mut self, value: &Self, condition: Condition) {
@@ -65,6 +93,12 @@ impl Cmov for u32 {
     }
 }
 
+impl CmovEq for u32 {
+    fn cmovne(&self, rhs: &Self, input: Condition, output: &mut Condition) {
+        csel_eq!("csel {2:w}, {3:w}, {4:w}, NE", self, rhs, input, output);
+    }
+}
+
 impl Cmov for u64 {
     #[inline(always)]
     fn cmovnz(&mut self, value: &Self, condition: Condition) {
@@ -86,5 +120,11 @@ impl Cmov for u64 {
             *value,
             condition
         );
+    }
+}
+
+impl CmovEq for u64 {
+    fn cmovne(&self, rhs: &Self, input: Condition, output: &mut Condition) {
+        csel_eq!("csel {2:w}, {3:w}, {4:w}, NE", self, rhs, input, output);
     }
 }
--- a/cmov/src/aarch64.rs
+++ b/cmov/src/aarch64.rs
@@ -2,10 +2,10 @@ use crate::{Cmov, CmovEq, Condition};
 use core::arch::asm;
 
 macro_rules! csel {
-    ($cmp:expr, $csel:expr, $dst:expr, $src:expr, $condition:expr) => {
+    ($csel:expr, $dst:expr, $src:expr, $condition:expr) => {
         unsafe {
             asm! {
-                $cmp,
+                "cmp {0:w}, 0",
                 $csel,
                 in(reg) $condition,
                 inlateout(reg) *$dst,
@@ -42,24 +42,12 @@ macro_rules! csel_eq {
 impl Cmov for u16 {
     #[inline(always)]
     fn cmovnz(&mut self, value: &Self, condition: Condition) {
-        csel!(
-            "cmp {0:w}, 0",
-            "csel {1:w}, {2:w}, {3:w}, NE",
-            self,
-            *value,
-            condition
-        );
+        csel!("csel {1:w}, {2:w}, {3:w}, NE", self, *value, condition);
     }
 
     #[inline(always)]
     fn cmovz(&mut self, value: &Self, condition: Condition) {
-        csel!(
-            "cmp {0:w}, 0",
-            "csel {1:w}, {2:w}, {3:w}, EQ",
-            self,
-            *value,
-            condition
-        );
+        csel!("csel {1:w}, {2:w}, {3:w}, EQ", self, *value, condition);
     }
 }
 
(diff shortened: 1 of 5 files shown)


tarcieri:
Looks great at first glance. I can take a more detailed look later.

> I believe the compiler will still attempt to optimise in some places, although I'm not 100% sure on that. Using this crate within subtle (or similar) sounds the most suitable to me.

The portable implementation could borrow `subtle`'s trick of using a volatile write as an optimization barrier.

Though I agree finding some way to get all of this work upstream into `subtle` seems good.

brxken128:
> Looks great at first glance. I can take a more detailed look later.

Thank you! I'll take a look into the volatile write, and also take a look at how `subtle` is structured, primarily to see how large of a task it'd be.

I don't assume it'd be overly difficult, and it seems very beneficial so I'm all for it.

brxken128:
I have some very promising news!

Using the `x86` implementation in the PR, along with `dudect_bencher`, here are the results for `cmoveq` and `cmovne` (with 100 million measurements):

```
running 10 benches
bench cmoveq_u128 seeded with 0xc2eac874bbfff9f8
bench cmoveq_u128 ... : n == +100.000M, max t = -0.76411, max tau = -0.00008, (5/tau)^2 = 4281830469
bench cmoveq_u16  seeded with 0x1d448ef882317e00
bench cmoveq_u16  ... : n == +100.000M, max t = +1.25790, max tau = +0.00013, (5/tau)^2 = 1579961129
bench cmoveq_u32  seeded with 0x616eabe391b02a50
bench cmoveq_u32  ... : n == +100.000M, max t = -1.85403, max tau = -0.00019, (5/tau)^2 = 727287432
bench cmoveq_u64  seeded with 0x944c16dd3a738d52
bench cmoveq_u64  ... : n == +100.000M, max t = -2.05989, max tau = -0.00021, (5/tau)^2 = 589188106
bench cmoveq_u8   seeded with 0xeae6f3b6361779ca
bench cmoveq_u8   ... : n == +99.889M, max t = -0.87249, max tau = -0.00009, (5/tau)^2 = 3280448987
bench cmovne_u128 seeded with 0x1b08ea492f9608de
bench cmovne_u128 ... : n == +99.568M, max t = +1.84216, max tau = +0.00018, (5/tau)^2 = 733503522
bench cmovne_u16  seeded with 0x48573db461a7b65b
bench cmovne_u16  ... : n == +100.000M, max t = +0.70990, max tau = +0.00007, (5/tau)^2 = 4960715094
bench cmovne_u32  seeded with 0x98279b1b4c577fed
bench cmovne_u32  ... : n == +100.000M, max t = +1.49098, max tau = +0.00015, (5/tau)^2 = 1124598322
bench cmovne_u64  seeded with 0x700c4d546f7f329d
bench cmovne_u64  ... : n == +99.831M, max t = +1.11933, max tau = +0.00011, (5/tau)^2 = 1991996799
bench cmovne_u8   seeded with 0xd04b925e2035c828
bench cmovne_u8   ... : n == +38.389M, max t = +1.71178, max tau = +0.00028, (5/tau)^2 = 327532597
```
[Here's the benchmark](https://gist.github.com/brxken128/7c7bbd42a79d0288addd89b8b5f95a6f). Obviously my methodology could be flawed, or `dudect_bencher` itself could be, but I don't yet have a reason to doubt the accuracy.

I'm unsure why `cmovne_u8` didn't complete the full cycle - I will re-run it and include it here in a moment.

tarcieri:
FWIW Intel guarantees the CMOV family of instructions is constant-time on all extant CPUs, and the use of inline assembly should ensure LLVM does not alter the emitted assembly code

brxken128:
> FWIW Intel guarantees the CMOV family of instructions is constant-time on all extant CPUs, and the use of inline assembly should ensure LLVM does not alter the emitted assembly code

I recall seeing it in the crate docs, but it never hurts to validate the implementation I suppose (even if just a sanity check). I wonder how this will act on `Aarch64`.

> The portable implementation could borrow subtle's trick of using a volatile write as an optimization barrier.

Which part of the portable implementation were you referring to here? I had a look but couldn't quite find where it'd fit in.

On another note, I ran the benches using the portable backend, and it also looks promising. It's not collecting the full 100M measurements here either - I'm not too sure what's going on, but I assume there's enough to be reliable.

```
running 10 benches
bench cmoveq_u128 seeded with 0x395c16b9e516050e
bench cmoveq_u128 ... : n == +100.000M, max t = +1.55722, max tau = +0.00016, (5/tau)^2 = 1030960941
bench cmoveq_u16  seeded with 0x8d3f1cb6f23e4d46
bench cmoveq_u16  ... : n == +40.532M, max t = -1.28140, max tau = -0.00020, (5/tau)^2 = 617124023
bench cmoveq_u32  seeded with 0xf0c6688cbbebefd5
bench cmoveq_u32  ... : n == +41.231M, max t = +1.26486, max tau = +0.00020, (5/tau)^2 = 644286082
bench cmoveq_u64  seeded with 0x968a9fa0ea121e3b
bench cmoveq_u64  ... : n == +39.363M, max t = +0.34390, max tau = +0.00005, (5/tau)^2 = 8320645836
bench cmoveq_u8   seeded with 0x47c5285a05d1159e
bench cmoveq_u8   ... : n == +100.000M, max t = -2.00943, max tau = -0.00020, (5/tau)^2 = 619149459
bench cmovne_u128 seeded with 0xbf87b34e478c30ce
bench cmovne_u128 ... : n == +100.000M, max t = +1.18042, max tau = +0.00012, (5/tau)^2 = 1794170662
bench cmovne_u16  seeded with 0x8c8925270439a99b
bench cmovne_u16  ... : n == +100.000M, max t = -1.31689, max tau = -0.00013, (5/tau)^2 = 1441593334
bench cmovne_u32  seeded with 0x77bc338505a0dcb9
bench cmovne_u32  ... : n == +100.000M, max t = +0.63647, max tau = +0.00006, (5/tau)^2 = 6171382062
bench cmovne_u64  seeded with 0xda022ec5135c349b
bench cmovne_u64  ... : n == +41.537M, max t = +1.01576, max tau = +0.00016, (5/tau)^2 = 1006469926
bench cmovne_u8   seeded with 0xbf6e437458bd8a00
bench cmovne_u8   ... : n == +100.000M, max t = -1.27892, max tau = -0.00013, (5/tau)^2 = 1528455770
```

tarcieri:
@brxken128 is there a minimal repro of that? I'd like to take a look in Godbolt

brxken128:
> @brxken128 is there a minimal repro of that? I'd like to take a look in Godbolt

This should do it:

```
macro_rules! cmov_eq {
    ($instruction:expr, $lhs:expr, $rhs:expr, $condition:expr, $dst:expr) => {
        let mut tmp = *$dst as u16;
        unsafe {
            asm! {
                "xor {0:e}, {1:e}",
                $instruction,
                in(reg) *$lhs,
                in(reg) *$rhs,
                inlateout(reg) tmp,
                in(reg) $condition as u16,
                options(pure, nomem, nostack),
            };
        };
        *$dst = tmp as u8;
    };
}

pub fn cmovne_u32(src: &u32, rhs: &u32, input: u8, output: &mut u8) {
    cmov_eq!("cmovnz {2:e}, {3:e}", src, rhs, input, output);
}

fn main() {
    let mut dst = 1u8;
    cmovne_u32(&1, &2, 0u8, &mut dst);
}
```

Inlining `cmovne_u32` seems to break (at least for me locally), although I'm not quite sure why.

tarcieri:
@brxken128 I went through in godbolt and didn't notice anything broken in particular. It looked like it inlined the `cmovne_u32` call in `main` regardless of what inlining hint was suggested.

What's failing exactly?

brxken128:
> What's failing exactly?

I had `cargo watch` running while I was code-editing (to re-test on every save), and this was active while I started adding `#[inline(always)]` to all of the functions.

Once I got to the `x86`, tests started failing. It only happens on some - others just run infinitely. It's really spontaneous to be honest.

Take this impl from `x86`:

```rust
impl CmovEq for u32 {
    fn cmovne(&self, rhs: &Self, input: Condition, output: &mut Condition) {
        cmov_eq!("cmovnz {2:e}, {3:e}", self, rhs, input, output);
    }
}
```

This works great, until `#[inline(always)]` is added above the function, and then the relevant test fails.

I'm unsure if this failing is down to an error in my code, an error with Rust, or an error with even my testing methodology but I don't think inline-always should be causing something such as this - everything works great without it.

tarcieri:
It'd be good to get the whole test into godbolt so you can see the difference in the emitted assembly

brxken128:
> It'd be good to get the whole test into godbolt so you can see the difference in the emitted assembly

[It runs fine](https://godbolt.org/z/55j4bMzT7) in godbolt, but not locally. Extremely odd.

I compared the two ASM outputs and don't see anything awry.

I'll commit the changes, push and see if CI shows similar failures.

tarcieri:
`cmp` is fine and won't branch. Behind the scenes it's actually a `sub` instruction.

If you can just compose it now in terms of the existing CMOV instructions (i.e. not define any additional assembly), doing XOR in Rust and using e.g. `Cmov::cmovz` on the result, that should work fine for now.

We could investigate doing optimized assembly as a separate pass.

brxken128:
Done as of my latest commit - I'll keep the ASM for `Aarch64` as it seems to work perfectly well from my testing.

tarcieri (approved):
Looks good now, thanks!

PR #885, opened 2023-04-02
Use minimal permissions for CI jobs

newpavlov:
For most jobs we only need to read the contents to run tests. But I am not sure which permissions should be used for the security audit job. I think it should be something like:
```yaml
permissions:
  contents: read
  checks: write
  issues: write
```
For now I will leave the default permissions.

Relevant issues: https://github.com/actions-rs/audit-check/issues/218, https://github.com/actions-rs/audit-check/issues/220

### DIFF
--- a/.github/workflows/blobby.yml
+++ b/.github/workflows/blobby.yml
@@ -8,6 +8,9 @@ on:
   push:
     branches: master
 
+permissions:
+  contents: read
+
 defaults:
   run:
     working-directory: blobby
--- a/.github/workflows/block-buffer.yml
+++ b/.github/workflows/block-buffer.yml
@@ -8,6 +8,9 @@ on:
   push:
     branches: master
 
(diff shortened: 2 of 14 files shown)


tarcieri:
FWIW I went through the repos and set the permissions to read-only quite awhile ago:

https://github.com/RustCrypto/utils/settings/actions

Screenshot 2023-04-02 at 8 31 22 AM

This could be belt-and-suspenders, I guess, but using administrative actions rather than configurations seems like a bit more surefire approach.

newpavlov:
It looks like this setting is equivalent to the [restricted column](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) with some additional rights to manipulate pull requests. It will be hard to do something malicious with such rights, but either way using tighter permissions probably will not hurt.

PR #900, opened 2023-05-18
Zeroize: Implement Zeroize for MaybeUninit, and remove some unsafe

elichai:
This PR contains a few things:
1. Implements Zeroize for `MaybeUninit<Z>` for any `Z`.
2.  Uses a const to create the `1` in   `impl_zeroize_for_non_zero` instead of using `unsafe` there.
3. For `Vec<T>` uses `spare_capacity_mut()` instead of manually creating a slice via unsafe. (This bumps MSRV to 1.60, So I understand if it's not wanted right now)
4. Uses the existing macros to generate more implementations.
5. On `CString` use `into_bytes_with_nul` instead of `into_bytes` as the latter pops the null character for no reason.
6. Use the `ptr::cast` method to cast types

If anything here is controversial I can drop some of the commits and/or split this into multiple PRs.

### DIFF
--- a/.github/workflows/zeroize.yml
+++ b/.github/workflows/zeroize.yml
@@ -26,7 +26,7 @@ jobs:
     strategy:
       matrix:
         rust:
-          - 1.56.0 # MSRV
+          - 1.60.0 # MSRV
           - stable
         target:
           - armv7a-none-eabi
@@ -53,7 +53,7 @@ jobs:
           # 32-bit Linux
           - target: i686-unknown-linux-gnu
             platform: ubuntu-latest
-            rust: 1.56.0 # MSRV
+            rust: 1.60.0 # MSRV
             deps: sudo apt update && sudo apt install gcc-multilib
           - target: i686-unknown-linux-gnu
             platform: ubuntu-latest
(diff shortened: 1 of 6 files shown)


tarcieri:
Looks like this needs an MSRV bump to 1.57, which I'd be OK with.

Otherwise it generally looks good.

tarcieri:
```
error[E0658]: use of unstable library feature 'vec_spare_capacity'
   --> zeroize\src\lib.rs:560:14
    |
560 |         self.spare_capacity_mut().zeroize();
    |              ^^^^^^^^^^^^^^^^^^
    |
    = note: see issue #75017 <https://github.com/rust-lang/rust/issues/75017> for more information
```

Make that MSRV 1.60.

elichai:
> ```
> error[E0658]: use of unstable library feature 'vec_spare_capacity'
>    --> zeroize\src\lib.rs:560:14
>     |
> 560 |         self.spare_capacity_mut().zeroize();
>     |              ^^^^^^^^^^^^^^^^^^
>     |
>     = note: see issue #75017 <https://github.com/rust-lang/rust/issues/75017> for more information
> ```
> 
> Make that MSRV 1.60.

Ops, fixed :)

tarcieri (approved):

PR #919, opened 2023-06-15
cpufeatures: check OS register support

### LINKED ISSUE #793: cpufeatures: add XGETBV checks
According to the [Intel manual](https://www.intel.com/content/dam/develop/external/us/en/documents/36945), when detecting features like AVX and AVX2 we should first check availability of the XGETBV instruction and then check that OS supports XMM and YMM registers.

newpavlov:
Fixes #793

cc @gabrielesvelto

### DIFF
--- a/cpufeatures/src/x86.rs
+++ b/cpufeatures/src/x86.rs
@@ -14,12 +14,12 @@ macro_rules! __unless_target_features {
     ($($tf:tt),+ => $body:expr ) => {{
         #[cfg(not(all($(target_feature=$tf,)*)))]
         {
-            #[cfg(not(any(target_env = "sgx", target_os = "none", target_os = "uefi")))]
+            #[cfg(not(any(target_env = "sgx", target_os = "", target_os = "uefi")))]
             $body
 
             // CPUID is not available on SGX. Freestanding and UEFI targets
             // do not support SIMD features with default compilation flags.
-            #[cfg(any(target_env = "sgx", target_os = "none", target_os = "uefi"))]
+            #[cfg(any(target_env = "sgx", target_os = "", target_os = "uefi"))]
             false
         }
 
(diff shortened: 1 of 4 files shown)


cpufeatures/src/x86.rs
```diff
@@ -61,18 +61,46 @@ macro_rules! __detect_target_features {
...
         #[doc(hidden)]
         macro_rules! check {
             $(
-                ($cr:expr, $name) => {
-                    true
+                ($cr:expr, $name) => {{
+                    let reg_cap = match $reg_cap {
+                        "xmm" => $crate::__xgetbv!($cr, 0b110),
+                        "ymm" => $crate::__xgetbv!($cr, 0b110),
+                        "zmm" => $crate::__xgetbv!($cr, 0b110),
```
newpavlov, line 96: Use correct mask values for XMM and ZMM.

cpufeatures/src/x86.rs
```diff
@@ -83,32 +111,32 @@ macro_rules! __expand_check_macro {
...
-    ("avx512pf", 1, ebx, 26),
-    ("avx512er", 1, ebx, 27),
-    ("avx512cd", 1, ebx, 28),
-    ("sha", 1, ebx, 29),
-    ("avx512bw", 1, ebx, 30),
-    ("avx512vl", 1, ebx, 31),
+    ("mmx", "none", 0, edx, 23),
+    ("sse", "xmm", 0, edx, 25),
+    ("sse2", "xmm", 0, edx, 26),
+    ("sse3", "xmm", 0, ecx, 0),
```
newpavlov, line 117: Double check this list.

tarcieri:
It looks like you're using `XGETBV` without first checking `CPUID` to determine if it's available.

I was reading about some virtualized environments that didn't implement it yesterday (although I forget specifically which ones, maybe I can dig up a link)

newpavlov:
I check it on line 76. Bit 26 of ECX register is responsible for availability of XSAVE, XRSTOR, XSETBV, XGETBV instructions.

newpavlov:
According to this [page](https://wiki.osdev.org/CPU_Registers_x86#Extended_Control_Registers), there are 3 bits relevant to AVX-512:
- 5 	`opmask` 	AVX-512 enabled and XSAVE support for opmask registers k0-k7
- 6 	`ZMM_Hi256` 	AVX-512 enabled and XSAVE support for upper halves of lower ZMM registers
- 7 	`Hi16_ZMM` 	AVX-512 enabled and XSAVE support for upper ZMM registers 

Honestly, I have no idea when we should check what. For now to be safe I will check all bits. We may change it in future according to user reports.

Issue #895, opened 2023-05-07
allow "0x" prefix


jordy25519:
Allow a "0x" prefix in the hex! literal, is there any appetite to support this?
find myself manually removing the prefix alot to satisfy the macro

newpavlov:
Can you show an example?

Usually, `0x` is used with hex-encoded integers and `hex-literal` probably should not be used in such cases. You either should use built-in integer literals (e.g. like we do [here](https://github.com/RustCrypto/hashes/blob/master/sha2/src/consts.rs)), or something like `U256::from_be_hex` defined in [`crypto-bigint`](https://docs.rs/crypto-bigint/latest/crypto_bigint/index.html).

jordy25519:
I frequently copy paste values like this as they are presented in the wild like so: (e.g. some address, transaction/block hash)
```rs
const value: [u8; 32] = hex!("0xec87ba8be686088e6248b03cc1de1afbc69cb043bfdd90af6d4fde8ed46979ea");
```

it's a minor thing :)
likely crypto-_currency_ specific but I figured the macro would have no issue supporting it as a ux improvement so wanted to see if it was acceptable

PR #923, opened 2023-06-26
hex-literal: add support for hex prefixes

DaniPopes:
Allows writing `hex!("0x1234")` by ignoring the initial `"0x"` prefix.
This only works by stripping the start of the literals, so something like `hex!("0x12 0x34")` or `hex!("   0x1234")` wouldn't work.
This would require a bit more effort to implement, so please let me know if you want this as well.

### DIFF
--- a/hex-literal/CHANGELOG.md
+++ b/hex-literal/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+### Added
+- Support for `0x` hex prefixes ([#923])
+
+[#923]: https://github.com/RustCrypto/utils/pull/923
+
 ## 0.4.1 (2023-04-05)
 ### Changed
 - Enforce const evaluation ([#889])
--- a/hex-literal/README.md
+++ b/hex-literal/README.md
@@ -10,15 +10,17 @@ This crate provides the `hex!` macro for converting hexadecimal string literals 
 
 It accepts the following characters in the input string:
 
+- `0x` — the hex prefix which is valid only at the start of a literal and will be ignored
 - `'0'...'9'`, `'a'...'f'`, `'A'...'F'` — hex characters which will be used in construction of the output byte array
 - `' '`, `'\r'`, `'\n'`, `'\t'` — formatting characters which will be ignored
 
 # Examples
+
 ```rust
 use hex_literal::hex;
 
 // The macro can be used in const contexts
-const DATA: [u8; 4] = hex!("01020304");
+const DATA: [u8; 4] = hex!("0x01020304");
 assert_eq!(DATA, [1, 2, 3, 4]);
 
 // Both upper and lower hex values are supported
--- a/hex-literal/src/lib.rs
+++ b/hex-literal/src/lib.rs
@@ -34,6 +34,18 @@ const fn next_byte(string: &[u8], pos: usize) -> Option<(u8, usize)> {
     Some(((half1 << 4) + half2, pos))
 }
 
+/// Strips the `0x` prefix from a hex string.
+///
+/// This function is an implementation detail and SHOULD NOT be called directly!
+#[doc(hidden)]
+pub const fn strip_hex_prefix(string: &[u8]) -> &[u8] {
+    if let [b'0', b'x' | b'X', rest @ ..] = string {
+        rest
+    } else {
+        string
+    }
+}
+
 /// Compute length of a byte array which will be decoded from the strings.
 ///
 /// This function is an implementation detail and SHOULD NOT be called directly!
@@ -80,7 +92,7 @@ pub const fn decode<const LEN: usize>(strings: &[&[u8]]) -> [u8; LEN] {
 #[macro_export]
 macro_rules! hex {
     ($($s:literal)*) => {{
-        const STRINGS: &[&'static [u8]] = &[$($s.as_bytes(),)*];
+        const STRINGS: &[&'static [u8]] = &[$( $crate::strip_hex_prefix($s.as_bytes()), )*];
         const LEN: usize = $crate::len(STRINGS);
         const RES: [u8; LEN] = $crate::decode(STRINGS);
         RES
--- a/hex-literal/tests/basic.rs
+++ b/hex-literal/tests/basic.rs
@@ -26,6 +26,12 @@ fn mixed_case() {
 }
 
 #[test]
+fn can_strip_prefix() {
+    assert_eq!(hex!("0x1a2b3c"), [0x1a, 0x2b, 0x3c]);
+    assert_eq!(hex!("0xa1" "0xb2" "0xc3"), [0xa1, 0xb2, 0xc3]);
+}
+
+#[test]
 fn multiple_literals() {
     assert_eq!(
         hex!(


newpavlov:
See #895. I don't think it's a good idea to support `0x` prefixes, since it can result in endianess-confusion. You should prefer using the [`crypto-bigint`](https://github.com/RustCrypto/crypto-bigint) crate when working with hex-encoded bigint values.

DaniPopes:
Makes sense for bigints, but this is just a wrapper around byte arrays so there should be no confusion IMO: `hex!("010203")` always is `[0x01, 0x02, 0x03]`
I think there should be the option to support this either way

newpavlov:
`0x`-prefix is commonly means big-endian order. When you write `0x01234567u32` it gets represented as `[0x67, 0x45, 0x23, 0x01]` on little-endian machines. So by encountering `hex!(0x01234567)` in code one may be quite surprised to find `[0x01, 0x23, 0x45, 0x67]` in memory.

Pzixel:
@newpavlov can you please consider revisiting this decision? The real-world application in my case for example is copying a lot of literals from other sources and using them as literals in the probram. In all cases these literals have 0x prefix, so need to tinker them manually in order to make macro parse it correctly is source of errors which isn't ideal. While I understand that from library perspective it may not be obvious but real-world usage shows that library would benefit from providing one way or another to express this. Maybe as an extra macro but I think it's very important to have it here.

tarcieri:
The macro does map the big endian representation of an integer to an array which has big endian byte ordering.

If we added support for hex literals with other types, I would expect it to use an architecture-specific byte ordering, the same way a `0x...` big endian literal maps to a host-specific integer.

newpavlov:
@Pzixel 
`hex-literal` is less than 100 lines long. IIUC you work with fixed size values, so you can write an even simpler `const fn` to do decoding for your project. It can look something like this:
```rust
pub const fn decode_u256(val: &str) -> [u8; 32] {
    let (prefix, bytes) = val.as_bytes().split_at(2);
    match prefix {
        b"0x" => (),
        _ => panic!("bad prefix"),
    }
    
    let mut i = 0;
    let mut buf = [0u8; 32];
    let mut buf_pos = 0;
    while i < bytes.len() {
        let mut pos = 0;
        while let Some((byte, new_pos)) = next_byte(bytes, pos) {
            buf[buf_pos] = byte;
            buf_pos += 1;
            pos = new_pos;
        }
        i += 1;
    }
    if buf.len() != buf_pos {
        panic!("length mismatch");
    }
    buf
}
```

Pzixel:
I'm affraid I'm not quite following why endianess is a subject to change here. From my perspective `0x` prefix just sets that the constant is written in hex, which is a bit exsessive for a library called `hed-literal` yet I don't see a harm in supporting it. Maybe I'm missing something?

> hex-literal is less than 100 lines long

While this true and I already used code from this PR I thought it's a good idea to try to clarify this for future library users so they maybe could have an easier time working with literals.

Proper bigint types are unfortunately lacking a const-time parsing (but they do have const conversion from arrays) which is a cornerstone here. And as I said I have implemented my own workarounds but they are a bit project-specific so I thought it's a good idea to have something in basic libs so everyone in community could benefit from it. But the main reason here is probably that I really don't understand how prefix implies ordering in any way

tarcieri:
> Proper bigint types are unfortunately lacking a const-time parsing

`crypto-bigint` supports constant-time parsing

yjhmelody:
Hi @newpavlov But rust support `0x` syntax for int. But you know that integer is always understood literally and has nothing to do with big or small endianness.

newpavlov:
@yjhmelody 
You said it yourself `0x` is used for integers. Endianness determines how those integers will be represented in memory. As I wrote above, it could makes sense to support `0x` for [`crypto_bigint::Uint`](https://docs.rs/crypto-bigint/latest/crypto_bigint/struct.Uint.html#method.from_le_hex). But the `hex!` macro creates a byte array, which is most certainly not an integer.

yjhmelody:
I think this is a syntax issue, and byte order is just the underlying representation, and for types such as slices and bytes, there is no byte order.

newpavlov:
I do not see your point. `0x` is traditionally used to indicate integer literals in using hex encoding in BE order. `hex!` is about byte arrays, not integers. If you code works with big integers, you should use `crypto-bigint` and its parsing utilities, not `hex-literal`.

I think we have properly articulated our opinion and there are no new points being provided, so it's unlikely we will change it.

yjhmelody:
Isn't it traditionally said that 0x is used to indicate that a value is hexadecimal? Why is it related to big endian order? can you give some links

tarcieri:
@yjhmelody `0x` is used as a sigil for describing an *integer literal* as hexadecimal bytes in big endian order, which is then converted into an integer in host endianness.

The `hex!` macro returns a byte array, not an integer. The literal syntax there is closer to `[0xAA, 0xBB, 0xCC, 0xDD]`.

`hex!` values are often interpreted as integers, but the endianness varies algorithm-by-algorithm. Older algorithms tend to use big endian; newer algorithms have in many cases shifted to little endian.

If this syntax were supported and wrote `u32::from_le_bytes(hex!("0xAABBCCDD"))`, it would decode to the value `0xDDCCBBAA` instead.

Pzixel:
I'm with @yjhmelody in this one, I'm not quite following. `0xAABBCCDD` should mean exactly `[0xAA, 0xBB, 0xCC, 0xDD]`. It's not an integer, it's an array, and user can decide on endianess when converting array to the value. The only thing macro should do is converting string into array properly. Bytes should come in the same order as they are in the string. 

if there are uses like `let x: i32 = hex!("0x11223344)` then I understand your position, but integers have native language support already and they don't need an extra crate while you can absultely legally write just `let x: i32 = 0x11223344`

tarcieri:
> 0xAABBCCDD should mean exactly [0xAA, 0xBB, 0xCC, 0xDD]. It's not an integer, it's an array

rustc would disagree with you:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=990f2cd9beef7875302a7b9e1c261bbd

```rust
assert_eq!(0xAABBCCDD, [0xAA, 0xBB, 0xCC, 0xDD])
```

```
error[E0277]: can't compare `{integer}` with `[{integer}; 4]`
 --> src/main.rs:2:5
  |
2 |     assert_eq!(0xAABBCCDD, [0xAA, 0xBB, 0xCC, 0xDD])
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no implementation for `{integer} == [{integer}; 4]`
  |
```

That said, the purpose of the `0x` prefix/sigil in an integer literal is to signify that follows is a base 16 integer.

But we already have the arguably even more clear `hex!` prefix on existing literals.

What does `hex!("0x...")` provide that `hex!("...")` does not? It's redundant.

yjhmelody:
> `u32::from_le_bytes(hex!("0xAABBCCDD"))`, it would decode to the value 0xDDCCBBAA instead.
it's `from_le_bytes` determining the repr not the hex syntax.

Pzixel:
> What does hex!("0x...") provide that hex!("...") does not? It's redundant.

It allows you to copy-paste existing hexadecimal value without modification from the other sources that provide them. For example, you can copy an `etherscan` address and use it in the program with no modifications. You might find this minor but this is actually a big deal. I was finding a lot of bugs because extra symbols got cut etc.

Its' redundant but it allows more easy integration with other existing places where it is used. 

> rustc would disagree with you:

I was saying about your macro that returns `[u8; _]`, it doesn't return an integer, so I'm not sure what this proves tbh. If we already return an array (and we do) we can only arrange bytes in this array, and since there is only one arrangement that makes sense I don't see how endianess comes into play here.

So tldr would be like: on the internet 0x just means that value is in hex. Often times we want to use such a value in our codebase. And we would like to avoid requirement to strip the prefix - because while in codebase hex! clearly means that underlying value is hex, on the internet it's just a number, so they always get shipped with it - and when copy-pasting it (because typing it manually is another source of error) you can accidentally introduce a bug, so "copy & go" approach would be nice even if it means some redundancy.

tarcieri:
> It allows you to copy-paste existing hexadecimal value without modification from the other sources that provide them.

But you have to add the `hex!("` on the front and `")` on the end anyway. Some modification is always required.

> I was saying about your macro that returns [u8; _], it doesn't return an integer, so I'm not sure what this proves tbh.

Integer literals with `0x` return an integer, not a byte array. That's the point.

> it's from_le_bytes determining the repr not the hex syntax.

No `from_le_bytes` is typically required with an integer literal, because it returns an integer, not bytes.

This is just going around in circles, so I'm going to lock it.

PR #925, opened 2023-07-05
cmov: XOR within the ASM block

### LINKED ISSUE #920: cmov: use ASM `XOR` for equality comparisons
This is how we originally did equality comparisons for x86, but for some reason `#[inline(always)]` caused it to break entirely.

I'd prefer to do the XOR within the ASM block as it's just more consistent, and I'll work on testing this out later today.

brxken128:
Now that #924 is a thing, we should be able to viably use `XOR` within the ASM block as opposed to within Rust.

This closes #920.

### DIFF
--- a/.github/workflows/workspace.yml
+++ b/.github/workflows/workspace.yml
@@ -20,7 +20,7 @@ jobs:
       - uses: RustCrypto/actions/cargo-cache@master
       - uses: dtolnay/rust-toolchain@master
         with:
-          toolchain: 1.67.0 # Pinned to prevent breakages
+          toolchain: 1.71.0 # Pinned to prevent breakages
           components: clippy
       - run: cargo clippy --all --all-features -- -D warnings
 
--- a/.github/workflows/zeroize.yml
+++ b/.github/workflows/zeroize.yml
@@ -26,7 +26,7 @@ jobs:
     strategy:
       matrix:
         rust:
-          - 1.56.0 # MSRV
+          - 1.60.0 # MSRV
           - stable
         target:
           - armv7a-none-eabi
@@ -53,7 +53,7 @@ jobs:
           # 32-bit Linux
           - target: i686-unknown-linux-gnu
             platform: ubuntu-latest
-            rust: 1.56.0 # MSRV
+            rust: 1.60.0 # MSRV
             deps: sudo apt update && sudo apt install gcc-multilib
           - target: i686-unknown-linux-gnu
             platform: ubuntu-latest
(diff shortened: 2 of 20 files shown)


brxken128:
This works locally on `x86_64-apple-darwin` and `x86_64-unknown-linux-gnu`.

tarcieri:
Looks like a legit test failure?

tarcieri:
The test failure seems legitimate?

brxken128:
> The test failure seems legitimate?

I'm not too sure. It's only happening for `u16` and `u128`, which is really odd. I've tried different register modifiers for all of the different types but will keep testing until I eventually get it working. I do believe u64 needs `:r` though. I had a thread asking about it in the Rust Discord server and someone extremely helpfully pointed out that `xor` does actually produce an output, so our second variable needed to be an `inout(reg) value => _` to be clobbered.

brxken128:
I tested locally on my R9 7950X, forcing it to use the `x86` module (by commenting the others out - apparently `miri` was enabled?` Anyway, the tests complete just fine.

I even temporarily removed `#[no_std]` so I could use `dbg!()` and ensure that the `x86` module was being used, and it indeed was. I wonder what's causing CI to continuously fail.

brxken128:
Further oddness - `cargo test (--release)` passes just fine, but `cargo nextest run (--release)` fails on:

```
FAIL [   0.003s] cmov::lib u128::cmoveq_works
FAIL [   0.003s] cmov::lib u16::cmoveq_works
FAIL [   0.003s] cmov::lib u16::cmovne_works
```

This was performed by forcing the x86 backend (by commenting the others out completely). AFAICT the code works just fine? Very odd.

brxken128:
Would it be possible for anyone else to give this a shot on `x86_64?` please?

Quick command: `git clone https://github.com/brxken128/utils utils-testing && cd utils-testing/cmov && git checkout x86-xor && cargo test` and then (if installed) try `cargo nextest run`. `Nexttest` seems to fail and I'm unsure why.

brxken128:
I fixed the issue thanks to a very helpful person in the Rust Discord server. It turns out that `:e` and 16 bit values don't mix when you're trying to compare the two, as `:e` is 32-bit and can/will contain garbage data in the upper-half.

This meant that the XOR was always inconsistent as we only wanted to compare the lower half of the register, so by using `:x` (which is 16 bit) we compare only the actual values, and not the "upper half", potentially garbage data.

I am very grateful for the Rust Discord server, and this has helped me learn a lot - and helped this pull request become merge-ready.

PR #926, opened 2023-07-05
cpufeatures: add support for `avx512vbmi(2)`

brxken128:
This has only been tested on my machine (R7 7950X), but the values seem correct to me - I was following [this chart](https://en.wikichip.org/wiki/x86/avx-512#Detection) and a PR from the other month.

### DIFF
--- a/cpufeatures/src/x86.rs
+++ b/cpufeatures/src/x86.rs
@@ -142,6 +142,6 @@ __expand_check_macro! {
     ("sha", "xmm", 1, ebx, 29),
     ("avx512bw", "zmm", 1, ebx, 30),
     ("avx512vl", "zmm", 1, ebx, 31),
-    ("avx512vbmi", 1, ecx, 1),
-    ("avx512vbmi2", 1, ecx, 6),
+    ("avx512vbmi", "zmm", 1, ecx, 1),
+    ("avx512vbmi2", "zmm," 1, ecx, 6),
 }
--- a/cpufeatures/src/x86.rs
+++ b/cpufeatures/src/x86.rs
@@ -142,4 +142,6 @@ __expand_check_macro! {
     ("sha", "xmm", 1, ebx, 29),
     ("avx512bw", "zmm", 1, ebx, 30),
     ("avx512vl", "zmm", 1, ebx, 31),
+    ("avx512vbmi", 1, ecx, 1),
+    ("avx512vbmi2", 1, ecx, 6)
 }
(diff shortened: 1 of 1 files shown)


cpufeatures/src/x86.rs
```diff
@@ -142,4 +142,6 @@ __expand_check_macro! {
     ("sha", "xmm", 1, ebx, 29),
     ("avx512bw", "zmm", 1, ebx, 30),
     ("avx512vl", "zmm", 1, ebx, 31),
+    ("avx512vbmi", 1, ecx, 1),
+    ("avx512vbmi2", 1, ecx, 6)
```
newpavlov, line 146: Add `"zmm"` after the feature names.

brxken128: After adding `"zmm"` and altering the tests to check for these specific features, they pass with flying colours.

`cpufeatures::new!(cpuid, "avx512vbmi", "avx512vbmi2");`

brxken128: Although I don't think that's the correct way to test? I'm unsure. I'll follow the docs and specifically test.

newpavlov: Check if the following assert passes on your machine:
```rust
cpufeatures::new!(cpuid, "avx512vbmi", "avx512vbmi2");

assert!(cpuid::get());
```

brxken128: I ran this test:

```rust
#[test]
fn l() {
    cpufeatures::new!(cpuid_avx512vbmi, "avx512vbmi", "avx512vbmi2");

    let token: cpuid_avx512vbmi::InitToken = cpuid_avx512vbmi::init();

    if token.get() {
        println!("CPU supports both");
    } else {
        println!("not supported");
    }
}
```

And it passed, but also changing `avx512vbmi` or `avx512vbmi2` to any character also passes the test.

brxken128: @newpavlov May I ask why the tests still pass, even with random characters in place? Is it due to the macro that defines the list?

newpavlov: Are you sure it passes with random feature name? As we can see [here](https://play.rust-lang.org/?gist=8b14ed974671593f749848f3bccdae42), the `new!` macro emits compilation error if an unknown feature is passed to it.

If you mean the `cpuid_avx512vbmi` part, then it's nothing more than a name of generated module. It can be any valid module name.

PR #930, opened 2023-07-06
Remove unnecessary cfgs on SIMD type impls

### LINKED ISSUE #929: Building for non-sse x86 (e.g. debian i386) with the zeroize feature fails.
While updating the debian package of the aes crate, I ran into a build failure on i386. Debian's i386 port (unlike the default configuration of rust upstream's i386 target) does not have sse2 enabled in it's default target.

building the aes crate with default features works fine, but if I enable the zeroize feature I get.

    error[E0277]: the trait bound `[core::arch::x86::__m128i; 11]: Zeroize` is not s                atisfied                                                                                         
       --> aes/src/ni.rs:195:43                                                                     
        |                                                                                            
    195 |                   zeroize::Zeroize::zeroize(&mut self.round_keys);                        
        |                   ------------------------- ^^^^^^^^^^^^^^^^^^^^ the trait                 `Zeroize` is not implemented for `[core::arch::x86::__m128i; 11]`                              
        |                   |                                                                       
        |                   required by a bound introduced by this call                             
    ...                                                                                         
    330 | / define_aes_impl!(                                                                       
    331 | |     Aes128,                                                                             
    332 | |     Aes128Enc,                                                                          
    333 | |     Aes128Dec,                                                                          
    ...   |                                                                                          
    338 | |     "AES-128",                                                                          
    339 | | );                                                                                      
        | |_- in this macro invocation                                                              
        |                                                                                           
        = help: the following other types implement trait `Zeroize`:                                
                  [MaybeUninit<Z>]                                                                   
                  [Z; N]                                                                             
                  [Z]                                                                                
        = note: this error originates in the macro `define_aes_impl` (in Nightly bui                lds, run with -Z macro-backtrace for more info)

newpavlov:
Closes #929

### DIFF
--- a/zeroize/src/aarch64.rs
+++ b/zeroize/src/aarch64.rs
@@ -8,13 +8,13 @@ use crate::{atomic_fence, volatile_write, Zeroize};
 use core::arch::aarch64::*;
 
 macro_rules! impl_zeroize_for_simd_register {
-    ($(($type:ty, $vdupq:ident)),+) => {
+    ($($type:ty),* $(,)?) => {
         $(
             #[cfg_attr(docsrs, doc(cfg(target_arch = "aarch64")))]
-            #[cfg_attr(docsrs, doc(cfg(target_feature = "neon")))]
             impl Zeroize for $type {
+                #[inline]
                 fn zeroize(&mut self) {
-                    volatile_write(self, unsafe { $vdupq(0) });
+                    volatile_write(self, unsafe { core::mem::zeroed() });
                     atomic_fence();
                 }
             }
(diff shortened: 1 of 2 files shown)


zeroize/src/aarch64.rs
```diff
@@ -8,13 +8,13 @@ use crate::{atomic_fence, volatile_write, Zeroize};
 use core::arch::aarch64::*;
 
 macro_rules! impl_zeroize_for_simd_register {
-    ($(($type:ty, $vdupq:ident)),+) => {
+    ($($type:ty),* $(,)?) => {
         $(
             #[cfg_attr(docsrs, doc(cfg(target_arch = "aarch64")))]
```
tarcieri, line 13: Looks like this can be removed too?

newpavlov: `uint8x8_t` and others are available only on AArch64 targets, so I don't think so.

tarcieri: I guess this crate hasn’t been moved to `doc_auto_cfg` yet

PR #934, opened 2023-08-20
block-padding: replace block iteration with fill()

tirz:
`slice.fill()` has been stabilized since rust 1.50 (`block-padding` currently require rust >= 1.56).


newpavlov:
Thanks!

PR #936, opened 2023-08-20
block-padding: implement ISO 10126

tirz:
This PR adds `rand_core` as a dependency for `block-padding`.
Maybe ISO 10126 should be put behind a feature?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -23,7 +23,15 @@ name = "block-padding"
 version = "0.3.3"
 dependencies = [
  "generic-array",
+ "rand_chacha",
+ "rand_core",
 ]
+
+[[package]]
+name = "cfg-if"
+version = "1.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


newpavlov:
We certainly do not want to depend on `getrandom`/`rand_core` by default. At the very least it should be behind a disabled-by-default feature.

Have you encountered the need for it in practice or is it done simply for completeness sake? IIRC random padding is usually frowned upon. So if it's the latter, I am not sure it's worth to add this (arguably) peculiar padding scheme.

tirz:
@newpavlov I added the feature `iso-10126`.
This is just for completeness sake.
I was working on a personal project and I decided to switch the Padding implementation to a generics P (Iso10126 is a possibility)... But really, I just need PKCS#7.

newpavlov:
Since ISO 10126 fits somewhat poorly into `block-padding`, I think it's better to simply omit it. If someone actually will need to use it in practice, then it can be implemented fairly easily outside of `block-padding`, our trait design allows to do it without much pain. Especially considering that the proposed implementation could be not flexible enough (e.g. it does not allow use of PRNG seeded with a fixed seed).

PR #944, opened 2023-09-05
Migrate to `hybrid-array`; MSRV 1.65

tarcieri:
Builds on RustCrypto/traits#1319.

Migrates the following crates away from using `generic-array` to using `hybrid-array` instead:

- `block-buffer`
- `block-padding`
- `dbl`
- `inout`

### DIFF
--- a/.github/workflows/block-buffer.yml
+++ b/.github/workflows/block-buffer.yml
@@ -25,7 +25,7 @@ jobs:
     strategy:
       matrix:
         rust:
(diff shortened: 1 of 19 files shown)


newpavlov (approved):

tarcieri:
Note that https://github.com/RustCrypto/traits/pull/1319 should probably land first, since `block-buffer` depends on `crypto-common`

PR #954, opened 2023-09-27
cmov: add ct_eq for slice comparison

codahale:
Add a simple `ct_eq` function for comparing two slices in constant time.


tarcieri:
I think you should be able to implement this generically as something like:

```rust
impl CmovEq for [T]
where
    T: CmovEq
{
   ...
}

codahale:
That’s a better idea. Done.

tarcieri (approved):

PR #958, opened 2023-10-13
hybrid-array: implement concat and split methods

rpiasetskyi:
These traits add convenient methods for concatenation and splitting. The size of the result is calculated in the compile time.

This implementation was inspired by `GenericArray` implementation of relevant traits.

This change is necessary for https://github.com/RustCrypto/traits/pull/1078


tarcieri:
Is there a particular reason to implement these in terms of traits instead of inherent methods? (other than that's what `generic-array` did)

It seems like since they're impl'd for `Array` universally, having them just be inherent methods would skip the step of having to import traits.

tarcieri (approved):

PR #966, opened 2023-10-22
zeroize: add more SIMD register support for aarch64

brxken128:
This PR adds SIMD support for: `int*`, `float*`, and `poly*` SIMD registers on `Aarch64`.

### DIFF
--- a/zeroize/src/aarch64.rs
+++ b/zeroize/src/aarch64.rs
@@ -1,7 +1,6 @@
 //! [`Zeroize`] impls for ARM64 SIMD registers.
 //!
-//! Gated behind the `aarch64` feature: MSRV 1.59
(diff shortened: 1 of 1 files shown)


zeroize/src/aarch64.rs
```diff
@@ -32,4 +30,22 @@ impl_zeroize_for_simd_register! {
...
+    float32x2_t,
+    float32x4_t,
+    float64x1_t,
+    float64x2_t,
+    poly8x8_t,
+    poly8x16_t,
+    poly16x4_t,
+    poly16x8_t,
+    poly64x1_t,
+    poly64x2_t,
```
tarcieri, line 50: I would suggest putting these in alphabetical order (I'd suggest, but they're all below `uint` already)

PR #967, opened 2023-10-22
zeroize: run clippy, clean up some docs and typos

brxken128:
This PR is rather self explanatory. I opted to allow `inline(always)` as in an application there's a good chance thst these will be pretty hot paths.

I also allowed wildcard imports within the `aarch64` and `x86` modules, as it'd just be a pain to keep track of all of the separate types.

Additionally, I chose to keep the notice about the lack of support for clearing registers. Inline-ASM is now a thing and end users are able to use it as they wish, but it's not specific to SIMD registers (which I failed to understand at first). We could maybe remove the part about the `rustc` support requirement/update it a little to make it more accurate for the current state of Rust, but I'm not really sure in that regard.

### DIFF
--- a/zeroize/src/aarch64.rs
+++ b/zeroize/src/aarch64.rs
@@ -5,6 +5,7 @@
 
 use crate::{atomic_fence, volatile_write, Zeroize};
 
+#[allow(clippy::wildcard_imports)]
 use core::arch::aarch64::*;
 
 macro_rules! impl_zeroize_for_simd_register {
--- a/zeroize/src/lib.rs
+++ b/zeroize/src/lib.rs
@@ -41,15 +41,13 @@
 //! ```
 //! use zeroize::Zeroize;
 //!
-//! fn main() {
-//!     // Protip: don't embed secrets in your source code.
-//!     // This is just an example.
-//!     let mut secret = b"Air shield password: 1,2,3,4,5".to_vec();
-//!     // [ ... ] open the air shield here
+//! // Protip: don't embed secrets in your source code.
+//! // This is just an example.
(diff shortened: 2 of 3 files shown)


tarcieri:
I'm not sure why this PR is needed. We already run `clippy` against all crates as part of the `workspace` workflow: https://github.com/RustCrypto/utils/blob/1dcf52ec901c01f24dc5c6d4b3df06cf773b6438/.github/workflows/workspace.yml#L16

Locally `clippy` passes for me with Rust 1.72 installed:

```console
$ rustc --version
rustc 1.72.0 (5680fa18f 2023-08-23)
$ cargo clippy --all-features -- -D warnings
warning: some crates are on edition 2021 which defaults to `resolver = "2"`, but virtual workspaces default to `resolver = "1"`
note: to keep the current resolver, specify `workspace.resolver = "1"` in the workspace root's manifest
note: to use the edition 2021 resolver, specify `workspace.resolver = "2"` in the workspace root's manifest
    Finished dev [optimized + debuginfo] target(s) in 0.00s
```

brxken128:
I have my flycheck configured with some (probably too harsh) settings, and these are just what were picked up while I was working on other parts of the crate - feel free to close this if you though as I can see why it's probably useless (although `size_of` was spelt `size_pf` in one spot, I could probably slip that singular change into one of my other open PRs though if preferable)

tarcieri:
I think clippy fixes should correspond to lints enabled in lib.rs, which also makes them reproducible (at least for a given version of clippy).

Since we're clean under the default configuration on the latest stable release of Rust, I'm going to close this.

tarcieri:
Also PRs for typos welcome

PR #968, opened 2023-10-22
zeroize: add support for `wasm`'s `v128` SIMD register

brxken128:
This PR adds support for zeroing `wasm`'s `v128` register. The module is named after the `core::arch` module, but it does require that `target_family = "wasm"` so I'm unsure of which naming scheme to stick with - I'm not too familiar with WASM as a whole to be honest.

I'll leave this as a draft until #967 is ready and merged, as the `support for zeroing SIMD registers for ...` notice will need updating.

### DIFF
--- a/zeroize/src/wasm32.rs
+++ b/zeroize/src/wasm32.rs
@@ -1,0 +1,22 @@
+//! [`Zeroize`] impls for WASM SIMD registers.
+
+use crate::{atomic_fence, volatile_write, Zeroize};
+
+use core::arch::wasm32::v128;
+
+macro_rules! impl_zeroize_for_simd_register {
+    ($($type:ty),* $(,)?) => {
+        $(
+            #[cfg_attr(docsrs, doc(cfg(target_arch = "wasm32", target_family = "wasm")))]
+            impl Zeroize for $type {
+                #[inline]
+                fn zeroize(&mut self) {
+                    volatile_write(self, unsafe { core::mem::zeroed() });
+                    atomic_fence();
+                }
+            }
+        )+
+    };
+}
+
+impl_zeroize_for_simd_register!(v128);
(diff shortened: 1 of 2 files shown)


zeroize/src/wasm32.rs
```diff
@@ -0,0 +1,22 @@
...
+                fn zeroize(&mut self) {
+                    volatile_write(self, unsafe { core::mem::zeroed() });
+                    atomic_fence();
+                }
+            }
+        )+
+    };
+}
+
+impl_zeroize_for_simd_register!(v128);
```
tarcieri, line 22: With only one register type, the macro is probably overkill, unless you plan to add more

brxken128: I agree, I was just adhering to the other implementations as I wasn't sure which style would be preferrential - bad decision on my part (even if it is just be abstracted away once compiled).

After looking further into it, `wasm` has only a generic `v128` type, and no other SIMD registers available (I believe the internal types are defined [here](https://github.com/llvm/llvm-project/blob/main/clang/lib/Headers/wasm_simd128.h) though).

I'll scrap the macro as it's definitely overkill for the single addressible register type, although I do wonder how the compiler even handles a type _this_ generic.

On another note, should there be platform-specific tests for (maybe all?) SIMD registers, so that we can validate that they're correctly zeroed out? I'm not too sure I've seen any within the codebase, but it might be good for additional validation/reassurance as even just this single type does seem rather internally-complex. I wouldn't mind giving it a go and opening a separate PR if I can get them working and ensure their correctness.

The tests might be unfeasable for many register types though, given that some are gated behind specific CPU features/extensions - I tried to add (feature gated) `avx512` register support, but that's both far too recent (MSRV of 1.73) and I couldn't even get it working on my 7950x, which has more than enough `avx512` support available. I probably spent way too long throwing flags at the compiler there haha

Issue #951, opened 2023-09-13
hybrid-array: Convert &[[T; N]] to &[Array<T, N>]


sinui0:
Noticed that yall are looking to migrate to `hybrid-array` instead of bumping to `generic-array` 1.0, thought I'd get ahead of this:

Can `hybrid-array` provide conversion methods to/from slices of core arrays?

Eg:

```rust
pub fn from_core_arrays_mut<const U: usize>(arrs: &mut [[T; U]]) -> &mut [Array<T, N>] {}
```

This helps support dependent code which already uses const generics.

Related: https://github.com/fizyk20/generic-array/issues/145

PR #974, opened 2023-10-27
hybrid-array: add slice_as_chunks(_mut) methods, minor refactoring

newpavlov:
As discussed [here](https://github.com/RustCrypto/traits/pull/1358#discussion_r1373973004) these functions will allow us to remove a bit of `unsafe` from downstream crates and it will be useful for users who would like to work with blocks of data.

### DIFF
--- a/hybrid-array/src/lib.rs
+++ b/hybrid-array/src/lib.rs
@@ -13,6 +13,7 @@
     clippy::cast_sign_loss,
     clippy::checked_conversions,
     clippy::implicit_saturating_sub,
+    clippy::arithmetic_side_effects,
     clippy::panic,
     clippy::panic_in_result_fn,
     clippy::unwrap_used,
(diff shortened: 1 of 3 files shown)


hybrid-array/src/lib.rs
```diff
@@ -13,7 +13,6 @@
     clippy::cast_sign_loss,
     clippy::checked_conversions,
     clippy::implicit_saturating_sub,
-    clippy::arithmetic_side_effects,
```
tarcieri, line 16: I'd prefer to keep this on at a crate-level. Perhaps you can disable it at the individual function level or operation level with a `SAFETY`-style comment about why overflow/underflow won't occur?

tarcieri:
#951 seems somewhat related, though not something I'm suggesting we necessarily tackle in this PR

tarcieri (approved):

newpavlov:
>#951 seems somewhat related, though not something I'm suggesting we necessarily tackle in this PR

Yeah, we can do it in a separate PR. I wonder if it can be handled by `From`/`Into` impls.

tarcieri:
@newpavlov that's the sort of line of thinking I was curious about, i.e. how to compose various types of (reference) conversions

PR #977, opened 2023-10-29
zeroize: Remove some unsafe code

joshlf:


newpavlov:
Personally, I don't think it's worth pulling `zerocopy` only for replacing `core::mem::zeroed()`.

joshlf:
Makes sense, no worries! Feel free to close.

PR #979, opened 2023-11-01
hybrid-array: Added serde impls for `Array`

rozbb:
I'd love to switch `hpke` over to `hybrid-array`, since it's becoming pretty hard to manually audit `generic-array`, and we expose it so it's not like we can get away with leaving it old forever.

One of the really convenient things that `generic-array` provides, though, is `Serialize`/`Deserialize` impls for `GenericArray`. We use this in our [Xyber impl](https://github.com/rozbb/rust-hpke/blob/6d3ec43f8b014a7d65faafc3b51d687b04087ce3/src/kem/xyber768d00.rs#L148). The only other option for autoderiving these traits is `serde-bigarray` (upstream support is [not](https://github.com/serde-rs/serde/pull/1860#pullrequestreview-735520420) coming any time soon), and I'd rather not bring it in.

This PR should bring `hybrid-array` a little closer to feature-parity with `generic-array`. I don't love the `unsafe` stuff here, but it's pretty minimal and well-documented, and also has a clear roadmap for removing it once certain functions stabilize.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,15 @@
 version = 3
 
 [[package]]
+name = "bincode"
+version = "1.3.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b1f45e9417d87227c7a56d22e471c6206462cba514c7590c09aff4cf6d1ddcad"
+dependencies = [
+ "serde",
+]
+
+[[package]]
 name = "blobby"
 version = "0.3.1"
 dependencies = [
@@ -84,6 +93,9 @@ version = "0.4.1"
 name = "hybrid-array"
 version = "0.2.0-pre.5"
 dependencies = [
+ "bincode",
+ "serde",
+ "serde_json",
  "typenum",
 ]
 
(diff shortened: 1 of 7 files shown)


rozbb:
It looks like some tests whose MSRV hasn't been bumped have a problem with  this `Cargo.toml`

tarcieri:
We can definitely bump MSRV on `fiat-constify`, FWIW.

rozbb:
I'm kind of in agreement here. I think we don't necessarily have to expect `[u8; N]` to be serializable. Or at the very least I can live without it and just write my own impls for my use case.

If you want to keep `ArrayExt::try_from_fn`, then I can remove the rest of this PR. Otherwise, feel free to close outright.

rozbb:
In this same vein. I thing maybe `hpke` shouldn't be doing serde impls either, for the same reason. It has `to_bytes` and `from_bytes`. I think that should be sufficient. If someone wants a specific format then they can choose.

tarcieri:
`hybrid-array` has moved to https://github.com/RustCrypto/hybrid-array

PR #984, opened 2023-11-18
hybrid-array: Impl Zeroize for Array

rozbb:
Let me know if you wanna feature gate this.

### DIFF
--- a/hybrid-array/Cargo.toml
+++ b/hybrid-array/Cargo.toml
@@ -22,4 +22,3 @@ zeroize = { version = "1.7", path = "../zeroize", optional = true }
 
 [features]
 default = []
-zeroize = ["dep:zeroize"]
--- a/hybrid-array/Cargo.toml
+++ b/hybrid-array/Cargo.toml
@@ -18,4 +18,8 @@ rust-version = "1.65"
 
 [dependencies]
 typenum = "1.17"
-zeroize = { version = "1.7", path = "../zeroize" }
+zeroize = { version = "1.7", path = "../zeroize", optional = true }
+
+[features]
+default = []
+zeroize = ["dep:zeroize"]
--- a/hybrid-array/Cargo.toml
+++ b/hybrid-array/Cargo.toml
@@ -18,3 +18,4 @@ rust-version = "1.65"
 
 [dependencies]
 typenum = "1.17"
+zeroize = { version = "1.7", path = "../zeroize" }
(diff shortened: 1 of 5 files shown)


rozbb:
I think it might not be possible to make `Drop` zeroize everything. If you try `impl<T: Zeroize, U: ArraySize> Drop for Array<T, U>`, it complains that you cannot specialize `Drop` impls.

hybrid-array/src/impls.rs
```diff
@@ -1,4 +1,22 @@
...
+        self.0.as_mut().iter_mut().zeroize()
+    }
+}
+
+impl<T, U> ZeroizeOnDrop for Array<T, U>
+where
+    T: ZeroizeOnDrop,
+    U: ArraySize,
+{
+}
```
tarcieri, line 19: A `zeroize` impl is good, but `Array` doesn't impl `Drop` and not every array needs to be zeroized on drop. Adding a `Drop` impl would also prevent moving the original array out of `Array`.

hybrid-array/Cargo.toml
```diff
@@ -18,3 +18,4 @@ rust-version = "1.65"
 
 [dependencies]
 typenum = "1.17"
+zeroize = { version = "1.7", path = "../zeroize" }
```
tarcieri, line 21: Yeah, feature-gating is good.

```suggestion
zeroize = { version = "1.7", optional = true, default-features = false, path = "../zeroize" }
```

rozbb: Should it be default?

hybrid-array/src/impls.rs
rozbb: I see. Array still is actually ZeroizeOnDrop tho right? Same reason `[Z; N]` is where `Z: ZeroizeOnDrop`

tarcieri: Aah sorry, that makes sense

hybrid-array/Cargo.toml
tarcieri: We have it feature-gated elsewhere, so making it off-by-default makes it easier to map to that, although you can always do `default-features = false`

rozbb:
Alright should be good to go

rozbb:
I believe build is failing because of outdated MSRV elsewhere

hybrid-array/Cargo.toml
```diff
@@ -18,3 +18,8 @@ rust-version = "1.65"
 
 [dependencies]
 typenum = "1.17"
+zeroize = { version = "1.7", path = "../zeroize", optional = true }
+
+[features]
+default = []
+zeroize = ["dep:zeroize"]
```
tarcieri, line 25: No need for this really

```suggestion
```

tarcieri:
#986 removed the MSRV from fiat-constify if you rebase

PR #985, opened 2023-11-19
algorithm-registry: initial commit

baloo:
This provides an helper for looking up an algorithm according to its OID.

I've found the same piece of code repeated in various ways all over in consumers of rust-crypto. This is an attempt at providing an implementation for rust-crypto users.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,17 @@
 version = 3
 
 [[package]]
+name = "algorithm-registry"
+version = "0.1.0"
+dependencies = [
+ "const-oid",
+ "digest",
+ "sha1",
+ "sha2",
+ "sha3",
+]
+
+[[package]]
 name = "blobby"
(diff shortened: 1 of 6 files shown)


baloo:
cc @raitobezarius

tarcieri:
This feels like the sort of thing that would belong in an omnibus crypto library, if we ever made one.

As is this is going to have all of the headaches of the dependency management for one and massive circular dependencies if you place it in this repo.

baloo:
I like the omnibus idea (and I like the name!). I think this would make a world of difference in the usability. I have no idea how to lay it (beside throwing a thousand helpers).

I'm not sure I see the point of the dependency management. I don't expect it to be too much of an issue, beside maybe duplicates in the Cargo.lock with local and remote/published (pulled from crates.io) versions.

In any case, let's close this for now.

PR #1016, opened 2023-12-31
hybrid-array: use `assert_ne!` in `slice_as_chunks(_mut)`

tarcieri:
Provides more effective debugging in the event the wrong size is used


tarcieri:
@newpavlov was there a specific reason these didn't use `assert_ne!` before?

PR #1039, opened 2024-01-10
hybrid-array: add `FromIterator` impl

tarcieri:
For feature parity with `GenericArray`.

It would be nice to have a fallible version of this which returns an error instead of panicking on length mismatch, but we'd first need to add `FromFn::try_from_fn` or thereabouts.

This is enough to cover the immediate use cases in the meantime.

### DIFF
--- a/hybrid-array/src/lib.rs
+++ b/hybrid-array/src/lib.rs
@@ -419,6 +419,25 @@ where
     }
 }
 
+impl<T, U> FromIterator<T> for Array<T, U>
+where
+    U: ArraySize,
+{
+    fn from_iter<I: IntoIterator<Item = T>>(iter: I) -> Self {
+        let mut iter = iter.into_iter();
+        let ret = Self::from_fn(|_| {
+            iter.next()
+                .expect("iterator should have enough items to fill array")
+        });
+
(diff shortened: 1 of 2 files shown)


hybrid-array/src/lib.rs
```diff
@@ -419,6 +419,25 @@ where
...
+    fn from_iter<I: IntoIterator<Item = T>>(iter: I) -> Self {
+        let mut iter = iter.into_iter();
+        let ret = Self::from_fn(|_| {
+            iter.next()
+                .expect("iterator should have enough items to fill array")
+        });
+
+        assert!(
+            iter.next().is_none(),
+            "too many items in iterator to fit in array"
```
tarcieri, line 435: Maybe these should both say "iterator should have exactly as many items as can fit in array"

newpavlov (approved):

hybrid-array/src/lib.rs
newpavlov: I think it could be useful to know whether you have more than needed or not enough elements in iterator to fill array. The other `expect` message can be "not enough items in iterator to fill array".

PR #1045, opened 2024-01-22
Adding a zeroize function for flat types

nstilt1:
For reducing zeroize code in chacha20, as per @newpavlov's [suggestion](https://github.com/RustCrypto/stream-ciphers/issues/336#issuecomment-1880346171)
I'm unsure of the name, documentation, and whether it should be a method instead of a function.

### DIFF
--- a/zeroize/src/lib.rs
+++ b/zeroize/src/lib.rs
@@ -795,7 +795,7 @@ unsafe fn volatile_set<T: Copy + Sized>(dst: *mut T, src: T, count: usize) {
     }
 }
 
-/// Zeroizes a flat type/struct. Only zeroizes the values that it owns, and it does not work on 
+/// Zeroizes a flat type/struct. Only zeroizes the values that it owns, and it does not work on
 /// dynamically sized values or trait objects.
 #[inline(always)]
 pub fn zeroize_flat_type<T: Sized>(data: &mut T) {
@@ -805,9 +805,7 @@ pub fn zeroize_flat_type<T: Sized>(data: &mut T) {
     //
     // This is safe because `mem::size_of<T>()` returns the exact size of the object in memory, and
     // `data_ptr` points directly to the first byte of the data.
-    unsafe {
-        volatile_set(data_ptr, 0, size)
-    }
+    unsafe { volatile_set(data_ptr, 0, size) }
     atomic_fence()
 }
 
--- a/zeroize/src/lib.rs
+++ b/zeroize/src/lib.rs
@@ -796,8 +796,8 @@ unsafe fn volatile_set<T: Copy + Sized>(dst: *mut T, src: T, count: usize) {
 }
 
 /// Zeroizes a flat type/struct. Only zeroizes the values that it owns, and it does not work on
-/// dynamically sized values or trait objects. Do not use this function on a type that already
-/// implements `ZeroizeOnDrop`.
+/// dynamically sized values or trait objects. It would be inefficient to use this function on a
+/// type that already implements `ZeroizeOnDrop`.
 ///
 /// # Safety
 /// - The type must not contain references to outside data or dynamically sized data, such as Vec<X>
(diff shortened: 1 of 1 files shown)


zeroize/src/lib.rs
```diff
@@ -795,6 +795,20 @@ unsafe fn volatile_set<T: Copy + Sized>(dst: *mut T, src: T, count: usize) {
     }
 }
 
+/// Zeroizes a flat type/struct. Only zeroizes the values that it owns, and it does not work on
+/// dynamically sized values or trait objects.
+#[inline(always)]
+pub fn zeroize_flat_type<T: Sized>(data: &mut T) {
```
newpavlov, line 801: The function must be `unsafe`, not only zeroized value may be invalid for `T`, but you also do not have any "flateness" guarantees.

newpavlov: It's also probably worth to make it accept `*mut T`.

nstilt1: It has been adjusted

zeroize/src/lib.rs
```diff
@@ -795,6 +795,25 @@ unsafe fn volatile_set<T: Copy + Sized>(dst: *mut T, src: T, count: usize) {
...
 }
 
+/// Zeroizes a flat type/struct. Only zeroizes the values that it owns, and it does not work on
+/// dynamically sized values or trait objects. Do not use this function on a type that already
+/// implements `ZeroizeOnDrop`.
+///
+/// # Safety
+/// - The type must not contain references to outside data or dynamically sized data
+/// - This function can invalidate the type if it is used after this function is called on it. It is
+///   advisable to call this method in `impl Drop`.
```
tarcieri, line 805: I think it would be good to add something here about requiring that the bit pattern of all zeros must be valid for the type

zeroize/src/lib.rs
```diff
@@ -795,6 +795,109 @@ unsafe fn volatile_set<T: Copy + Sized>(dst: *mut T, src: T, count: usize) {
...
+///
+/// // the String's metadata was overwritten (such as the pointer to the heap, length, capacity)
+/// assert_eq!(data.0.as_ptr().is_null(), true);
+/// assert_eq!(data.0.len(), 0);
+/// assert_eq!(data.0.capacity(), 0);
+///
+/// // the original data still lives on the heap, possibly resulting in a little memory leak
+/// let memory_inspection = unsafe { core::slice::from_raw_parts(original_ptr, 49) };
+/// assert_eq!(memory_inspection, b"take care when zeroizing with zeroize_flat_type()");
+/// ```
```
tarcieri, line 889: It might be better to list out the things this counter example is trying to demonstrate, e.g. as a bulleted list, rather than giving people working code.

The problem with giving people any bad code is they might copy/paste it without realizing it's intended to demonstrate what not to do.

zeroize/src/lib.rs
newpavlov: I do not think this requirement is needed for this function. We just need to mention that the function may leave the memory in an invalid state for type `T`.

In other words, I think (but not 100% sure) that the following code is safe:
```rust
#[repr(transparent)]
struct Foo(NonZeroU32);

impl Drop for Foo {
    fn drop(&mut self) {
        unsafe { zeroize_flat_type(self); }
    }
}
```

zeroize/src/lib.rs
nstilt1: I've made some revisions. I am not very knowledgeable about the ins and outs of the type system/rust (among other things), so some of the necessary information might be incorrect or missing. If you feel the need to change anything or add any important missing types or information, please feel free to update it.

tarcieri: Looks good now, thanks

zeroize/src/lib.rs
tarcieri: I'm not really clear on if that's allowed or not: https://github.com/rust-lang/unsafe-code-guidelines/issues/394

tarcieri: Notably it seems like it would be UB with `drop_in_place` or `ManuallyDrop::drop` which both permit access to the value after running the `Drop` impl

newpavlov: Both `drop_in_place` and `ManuallyDrop::drop` are `unsafe`, so it should not be a problem. Per ``ManuallyDrop::drop`` docs:

>However, this “zombie” value should not be exposed to safe code, and this function should not be called more than once. To use a value after it’s been dropped, or drop a value multiple times, can cause Undefined Behavior (depending on what drop does).

For now, we probably should write about _validity_ (with `NonZeroU32` as a counter-example), but mention that zeroized state is allowed to break _safety invariants_ of the type.

tarcieri: Sure, that sounds good

tarcieri (approved):

zeroize/src/lib.rs
nstilt1: Okay. Feel free to make any changes

PR #1053, opened 2024-02-26
feat(opaque-debug): support generics

sinui0:
This PR adds support for opaque debug impls on structs with generics. It does not support structs with trait bounds as working with them using macro_rules! is not possible as far as I can tell.

It also wasn't obvious how to avoid exporting the `format_params` macro which is required to build the format template without a trailing comma eg `Foo<T, U, V,> { ... }`.

### DIFF
--- a/opaque-debug/src/lib.rs
+++ b/opaque-debug/src/lib.rs
@@ -8,6 +8,17 @@
 #[doc(hidden)]
 pub extern crate core as __core;
 
+#[macro_export]
+#[doc(hidden)]
+macro_rules! format_params {
+    ($single:ident) => {
+        "{}"
+    };
+    ($first:ident, $($rest:ident),+) => {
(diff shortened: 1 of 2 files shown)


newpavlov:
We had some complaints about number of dependencies which our crates pull (despite that the pulled crates are part of the same org...) and we decided that pulling a whole crate to save in most cases [just 4 lines of code](https://github.com/RustCrypto/hashes/blob/master/sha1/src/lib.rs#L106-L110) is usually not worth the trouble.

>I do use this in some projects

In this case I think we can cut a new release.

newpavlov:
>I could trim to just the last element in the path based on the :: separator

But generic parameter may contain its own generic parameters, which will not be trimmed with a naive approach. :) I am fine with either option with a slight preference of using output `core::any::type_name` as-is, because it's a more straightforward and simple solution.

PR #1080, opened 2024-06-19
Fix Nightly warnings

newpavlov:


zeroize/src/lib.rs
```diff
@@ -409,10 +409,10 @@ where
         //
         // Safety:
         //
-        // The memory pointed to by `self` is valid for `mem::size_of::<Self>()` bytes.
+        // The memory pointed to by `self` is valid for `size_of::<Self>()` bytes.
         // It is also properly aligned, because `u8` has an alignment of `1`.
         unsafe {
-            volatile_set((self as *mut Self).cast::<u8>(), 0, mem::size_of::<Self>());
+            volatile_set((self as *mut Self).cast::<u8>(), 0, size_of::<Self>());
```
newpavlov, line 415: I am not sure why exactly compiler complains about `mem::size_of`. Was it added to the prelude?

tarcieri: Yeah, it was. Just encountered that here too: https://github.com/RustCrypto/traits/pull/1589

PR #1092, opened 2024-07-26
wycheproof2blb: secp256k1 with P1363 format

XuJiandong:
parsing ecdsa_secp256k1_sha256_p1363_test.json

### DIFF
--- a/wycheproof2blb/src/ecdsa.rs
+++ b/wycheproof2blb/src/ecdsa.rs
@@ -54,7 +54,7 @@ pub fn generator(data: &[u8], algorithm: &str, _key_size: u32) -> Vec<TestInfo> 
 
     let mut infos = vec![];
     for g in &suite.test_groups {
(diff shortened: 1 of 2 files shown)


tarcieri:
I guess these are vectors for the [IEEE P1363](https://en.wikipedia.org/wiki/IEEE_P1363), which unless it also goes by some other name (which I can't verify because I don't have access to the document) we don't support.

From Wikipedia:

> The IEEE Standard Association withdrew all of the 1363 standards except 1363.3-2013 on 7 November 2019.

XuJiandong:
The legacy IEEE 1363 has a very simple signature format with only r + s, totaling 64 bytes. I’ve searched some P1363 signatures in the test cases, and they are unique, not just copied from other cases.

tarcieri:
Aah, never heard it referred to that way. It's what we call just `ecdsa::Signature` or a "fixed" signature

tarcieri (approved):

PR #1100, opened 2024-08-24
cpufeatures: add support for DIT detection on Apple silicon

tarcieri:
Adds support for detecting Data Independent Timing (DIT) support which is available on Apple silicon (e.g. M1/M2/M3).

This uses the `hw.optional.arm.FEAT_DIT` sysctl for detection.

For more information, see:

https://developer.apple.com/documentation/xcode/writing-arm64-code-for-apple-platforms#Enable-DIT-for-constant-time-cryptographic-operations

### DIFF
--- a/cpufeatures/src/aarch64.rs
+++ b/cpufeatures/src/aarch64.rs
@@ -107,6 +107,12 @@ macro_rules! check {
     ("aes") => {
         true
     };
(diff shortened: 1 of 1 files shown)


tarcieri:
Note: I locally verified this was able to detect DIT on an M1

PR #1102, opened 2024-08-25
aarch64-dit: initial crate

tarcieri:
Adds a crate with wrappers for the Data-Independent Timing (DIT) feature of AArch64 CPUs.

The implementation is largely a translation of Apple's guide of how to write wrappers for enabling/disabling DIT: https://developer.apple.com/documentation/xcode/writing-arm64-code-for-apple-platforms#Enable-DIT-for-constant-time-cryptographic-operations

It would be nice to wrap that all up into an RAII guard which can first use `cpufeatures` to check for `FEAT_DIT` and, if available, enable it for the current thread, while also first querying the processor status register and restoring the previous state on `Drop`, which is necessary for proper nested usage of DIT.

But for now, this just wraps the barebones functionality in an `unsafe` API.

### DIFF
--- a/.github/workflows/aarch64-dit.yml
+++ b/.github/workflows/aarch64-dit.yml
@@ -24,7 +24,7 @@ jobs:
     strategy:
       matrix:
         toolchain:
-          - 1.60.0 # MSRV
+          - 1.61.0 # MSRV
           - stable
     runs-on: macos-latest
     steps:
--- a/.github/workflows/aarch64-dit.yml
+++ b/.github/workflows/aarch64-dit.yml
@@ -1,0 +1,37 @@
+name: aarch64-dit
+
+on:
+  pull_request:
+    paths:
+      - "aarch64-dit/**"
+      - "Cargo.*"
+  push:
+    branches: master
+
+permissions:
+  contents: read
+
+defaults:
+  run:
+    working-directory: aarch64-dit
+
+env:
+  CARGO_INCREMENTAL: 0
+  RUSTFLAGS: "-Dwarnings"
+
+jobs:
+  macos:
+    strategy:
+      matrix:
+        toolchain:
+          - 1.60.0 # MSRV
+          - stable
+    runs-on: macos-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: RustCrypto/actions/cargo-cache@master
+      - uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: ${{ matrix.toolchain }}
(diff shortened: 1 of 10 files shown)


.github/workflows/aarch64-dit.yml
```diff
@@ -0,0 +1,37 @@
...
+
+defaults:
+  run:
+    working-directory: aarch64-dit
+
+env:
+  CARGO_INCREMENTAL: 0
+  RUSTFLAGS: "-Dwarnings"
+
+jobs:
```
tarcieri, line 22: FWIW I tried to see if it would work under `cross` in Linux, but unfortunately not:

```
---- tests::get stdout ----
thread 'tests::get' panicked at aarch64-dit/src/lib.rs:46:13:
DIT is not available on this CPU
```

tarcieri:
Looks like this will need a higher MSRV:

```
error[E0658]: the target feature `dit` is currently unstable
  --> aarch64-dit/src/lib.rs:[15](https://github.com/RustCrypto/utils/actions/runs/10549036451/job/29223626727?pr=1102#step:5:16):18
   |
15 | #[target_feature(enable = "dit")]
   |                  ^^^^^^^^^^^^^^
   |
   = note: see issue #44839 <https://github.com/rust-lang/rust/issues/44839> for more information
```

aarch64-dit/src/lib.rs
```diff
@@ -0,0 +1,64 @@
...
+pub unsafe fn get_dit_enabled() -> bool {
+    let mut dit: u64;
+    asm!("mrs {dit}, DIT", dit = out(reg) dit);
+    (dit >> 24) & 1 != 0
+}
+
+/// Enable DIT for the current thread.
+#[target_feature(enable = "dit")]
+pub unsafe fn set_dit_enabled() {
+    asm!("msr DIT, #1");
```
newpavlov, line 25: I think it's worth to add proper `options` to this and other `asm!` blocks. I think `nomem`, `preserves_flags`, and `nostack` are applicable.

newpavlov (approved):

tarcieri:
At some point we might consider having an ISA-independent crate for this sort of instruction pattern, provided we can actually build a portable abstraction over it: https://www.intel.com/content/www/us/en/developer/articles/technical/software-security-guidance/best-practices/data-operand-independent-timing-isa-guidance.html

aarch64-dit/src/lib.rs
tarcieri: Added in 50a4d9e

PR #1108, opened 2024-09-06
aarch64-dit: safe, high-level API

tarcieri:
Removes the previous low-level unsafe API and wraps it up instead in a safe API which automatically performs runtime CPU feature detection and RAII guards for enabling DIT and restoring its previous state when done.

### DIFF
--- a/aarch64-dit/src/lib.rs
+++ b/aarch64-dit/src/lib.rs
@@ -6,14 +6,92 @@
 )]
 #![warn(missing_docs, rust_2018_idioms, unused_qualifications)]
 
+//! ## Usage
+//!
+//! ```
+//! use aarch64_dit::Dit;
+//!
+//! let dit = Dit::init();
+//! assert!(!dit.is_enabled());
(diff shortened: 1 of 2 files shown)


aarch64-dit/src/lib.rs
```diff
@@ -6,14 +6,92 @@
...
+    }
+
+    /// Enable Data-Independent Timing (if available).
+    ///
+    /// Returns an RAII guard that will return DIT to its previous state when dropped.
+    #[must_use]
+    pub fn enable(&self) -> Guard<'_> {
+        let was_enabled = if self.is_supported() {
+            let was_enabled = unsafe { get_dit_enabled() };
+            unsafe { set_dit_enabled() };
```
newpavlov, line 48: This will result in two separate function calls because `#[target_feature(enable = "dit")]` prevents inlining. It may be worth to tweak `set_dit_enabled` to return DIT state flag and write here `let was_enabled = unsafe { set_dit_enabled() };`

PR #1117, opened 2024-09-28
block-padding+inout: remove `std` feature

tarcieri:
Leverages the `error_in_core` feature stabilized in Rust 1.81 to allow `Error` trait impls in `no_std` contexts.

This was the only dependency on `std`, so this also removes the `std` feature, which is no longer necessary.

### DIFF
--- a/inout/Cargo.toml
+++ b/inout/Cargo.toml
@@ -14,7 +14,3 @@ readme = "README.md"
 [dependencies]
 block-padding = { version = "0.4.0-rc.0", path = "../block-padding", optional = true }
 hybrid-array = "0.2.0-rc.10"
-
-[package.metadata.docs.rs]
-all-features = true
-rustdoc-args = ["--cfg", "docsrs"]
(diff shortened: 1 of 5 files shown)


newpavlov (approved):
We also probably can remove the `[package.metadata.docs.rs]` section as well.

inout/Cargo.toml
```diff
@@ -14,10 +14,3 @@ readme = "README.md"
 [dependencies]
 block-padding = { version = "0.4.0-rc.0", path = "../block-padding", optional = true }
 hybrid-array = "0.2.0-rc.10"
-
-[features]
-std = ["block-padding/std"]
-
-[package.metadata.docs.rs]
-all-features = true
```
tarcieri, line 22: Looks like we still need the `all-features = true` here because of the `block-padding` feature.

I think `--cfg docsrs` might be set automatically now though?

newpavlov: >I think --cfg docsrs might be set automatically now though?

It would be great, if true. We should eventually remove it from other crates then.

PR #1131, opened 2024-11-11
Exclude `aarch64-dit` from root workspace

newpavlov:
Compiling the crate on non-AArch64 targets results in a compilation error, so having it in workspace can be problematic on other target arches.


tarcieri (approved):

tarcieri:
Hmm, perhaps you shouldn't include the Cargo.lock changes?

PR #1139, opened 2024-11-23
opaque-debug: Fix typo in example code

tyilo:


newpavlov:
Thank you! But note that `opaque-debug` is effectively deprecated and we instead recommend to implement "opaque" debugs manually.

tyilo:
Thanks, I'm not using it, I just randomly read the code :wink:

PR #1163, opened 2025-02-24
Update year in license files

newpavlov:


aarch64-dit/LICENSE-MIT
```diff
@@ -1,21 +1,25 @@
-MIT License
+Copyright (c) 2024-2025 The RustCrypto Project Developers
```
newpavlov, line 1: @tarcieri 
Note that I changed this file to the same template used in most our other crates. If you think that it's better to use the old variant, I think we need to update all crates.

tarcieri: seems fine

tarcieri (approved):

PR #1180, opened 2025-05-05
Add unsafe test to check that dropped values are zeroized

dvdplm:
Test that zeroization actually occurs when a boxed value goes out of scope.

Uses unsafe code, likely relies on UB and adds a custom allocator, hence the draft: is something like this acceptable?

c.f. https://github.com/iqlusioninc/crates/pull/1310


zeroize/tests/zeroize.rs
```diff
@@ -208,3 +208,60 @@ fn asref() {
...
+                ptr = *boxptr;
+                assert!(!ptr.is_null(), "ptr is null before drop, not ok");
+                let bytes: &[u8] = core::slice::from_raw_parts(ptr as *const u8, size_of::<u128>());
+                assert!(
+                    !bytes.iter().all(|&b| b == 0),
+                    "Expected non-zero data, instead found 0s: {:X?}",
+                    bytes
+                );
+            }
+            // Check that the memory is cleared after the scope ends
```
tarcieri, line 259: You can use [`drop_in_place`](https://doc.rust-lang.org/std/ptr/fn.drop_in_place.html) to invoke the drop handler (see other tests in the same module)

zeroize/Cargo.toml
```diff
@@ -27,9 +27,11 @@ default = ["alloc"]
 alloc = []
 std = ["alloc"]
 
-aarch64 = [] # NOTE: vestigial no-op feature; AArch64 support is always enabled now
+aarch64 = []                # NOTE: vestigial no-op feature; AArch64 support is always enabled now
 derive = ["zeroize_derive"]
-simd = [] # NOTE: MSRV 1.72
+simd = []                   # NOTE: MSRV 1.72
+
+test-allocator = ["std"]
```
newpavlov, line 34: I don't think you need `std` for this.

zeroize/tests/zeroize.rs
```diff
@@ -208,3 +208,60 @@ fn asref() {
     let _asmut: &mut [u8] = buffer.as_mut();
     let _asref: &[u8] = buffer.as_ref();
 }
+
+#[cfg(not(miri))]
+#[cfg(feature = "test-allocator")]
+mod zeroization_with_custom_allocator {
```
newpavlov, line 214: It maybe be worth to move this module into a separate test file, so the custom allocator would not affect the other tests.

zeroize/Cargo.toml
tarcieri: There shouldn't be any need for an additional crate feature at all, IMO

dvdplm: `std` is needed because of the `use std::alloc::{GlobalAlloc, Layout, System};` – the custom allocator falls back to the system allocator.

zeroize/tests/zeroize.rs
dvdplm: Sadly the `#[global_allocator]`, once enabled, does affect all tests. Until the `allocator_api` lands I know of no way around this. :/

Happy to move it to separate file though. :)

zeroize/tests/zeroize.rs
dvdplm: I saw those tests but felt that, given how magic `drop_in_place` is, readers would likely be mroe familiar with a plain old scope ending. Do you disagree?

zeroize/Cargo.toml
newpavlov: You don't need `std` for this. The same functionality is available in the `alloc` crate.

But considering that tests are built with `std` by default, it does not matter either way. You can simply remove the feature.

dvdplm: …and the feature is there to stop `#[global_allocator]` from kicking in for all tests all the time.

zeroize/tests/zeroize.rs
tarcieri: You can use a `cfg` attribute rather than a feature. Or, use a crate like `version_check` to detect nightly

zeroize/Cargo.toml
dvdplm: > You don't need `std` for this. The same functionality is available in the `alloc` crate.

You are correct, unless you need the system allocator, `System`. 

That's only available in `std::alloc` (compare alloc crate [here](https://doc.rust-lang.org/alloc/alloc/index.html#structs), std::alloc [here](https://doc.rust-lang.org/std/alloc/index.html#structs))

zeroize/tests/zeroize.rs
dvdplm: @tarcieri If using the nightly unstable API is acceptable (and preferable) that'd make it a bit less ugly. Is that your preference? 

I did not dig too deep in that particular rabbit hole but I think that the feature is being rewritten and is not close to stabilization.

tarcieri: My main concern is not introducing features for non-user-facing things like tests

dvdplm:
I guess the original question still stands: is something like this acceptable? I'm on the fence myself, hence the draft.

newpavlov:
Personally, while I am fine with having such test, I don't see much worth in it. I will leave the final decision to @tarcieri.

As a small suggestion, it may be worth to slightly modify how the test is implemented. Instead of reading leaked memory in the test itself, it may be better to allocate a known size (e.g. 1024 bytes) using `SecretBox` and then in `dealloc` check that deallocated memory with such size should be filled with zeros.

newpavlov:
@tarcieri WDYT? I lean towards closing this PR.

tarcieri:
While it would be nice to have a test, I worry about including one that has UB

newpavlov:
Closing in favor of #1199

PR #1187, opened 2025-06-23
blobby: replace iterators with `const fn` parsing

newpavlov:
The new approach parses blb files into statics which allows to use the crate in `no_std` contexts. This PR additionally introduces the `parse_into_structs` macro which defines a struct with byte slice fields and transforms blobby data into list of these structs.

For example, it allows us to write:
```rust
#[derive(Copy, Clone, Debug)]
struct TestVector {
    key: &'static [u8],
    nonce: &'static [u8],
    aad: &'static [u8],
    plaintext: &'static [u8],
    ciphertext: &'static [u8]
}

blobby::parse_into_structs!(
    include_bytes!("/path/to/file.blb");
    static TEST_VECTORS: &[
        TestVector { key, nonce, aad, plaintext, ciphertext }
    ];
);

// Expands into:
static TEST_VECTORS: &[TestVector] = { ... };
```

I also thought about adding support for byte array fields, but with the current Rust capabilities (i.e. without const traits) the resulting code becomes annoyingly complex, so I decided against it in the current version.

### DIFF
--- a/blobby/src/decode.rs
+++ b/blobby/src/decode.rs
@@ -215,10 +215,10 @@ macro_rules! parse_into_slice {
 macro_rules! parse_into_structs {
     (
         $data:expr;
-        $static_vis:vis static $items_name:ident;
-        $ty_vis:vis struct $item:ident {
-            $($field:ident),* $(,)?
-        }
+        #[define_struct]
+        $static_vis:vis static $items_name:ident: &[
+            $ty_vis:vis $item:ident { $($field:ident),* $(,)? }
+        ];
     ) => {
         #[derive(Debug, Clone, Copy, Eq, PartialEq)]
         $ty_vis struct $item {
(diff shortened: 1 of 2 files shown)


blobby/src/decode.rs
```diff
@@ -0,0 +1,271 @@
...
+            Err(err) => return Err(err),
+        }
+    };
+}
+
+pub const fn parse_dedup_len(mut data: &[u8]) -> Result<usize, Error> {
+    read_vlq(&mut data)
+}
+
+pub const fn parse_items_len(mut data: &[u8]) -> Result<usize, Error> {
```
newpavlov, line 57: Unfortunately, to get number of items stored in a blb file we have to fully parse it. In other words, `parse_into_slice` effectively parses the file twice, which may be noticeably slow on large enough files. We could resolve this by adding number of stored items to the start of the file, but it would mean that we have to re-generate our existing blb files.

PR #1194, opened 2025-08-04
dbl v0.4.0

### LINKED ISSUE #1077: dbl: retire crate and vendor implementation into downstream users?
The `dbl` crate contains an implementation of doubling / multiply-by-x for the polynomial x^128 + x^7 + x^2 + x + 1 over GF(2^128).

It's used in `aes-siv`, `cmac`, `pmac`, and `ocb3`. The same polynomial is used by GHASH, although the `ghash` crate doesn't use it because it's implemented in terms of POLYVAL, which uses the "reverse" polynomial.

The question is does it deserve its own crate, or is it trivial enough to simply be vendored as needed?

It might make more sense if we provided e.g. SIMD implementations which operate in parallel over multiple blocks.

tarcieri:
### Changed
- Migrated from `generic-array` to `hybrid-array` ([#944])
- Edition changed to 2024 and MSRV bumped to 1.85 ([#1149])
- Seal the `Dbl` trait ([#1198])

Closes #1077 

[#944]: https://github.com/RustCrypto/utils/pull/944
[#1149]: https://github.com/RustCrypto/utils/pull/1149
[#1198]: https://github.com/RustCrypto/utils/pull/1198

### DIFF
--- a/dbl/README.md
+++ b/dbl/README.md
@@ -26,7 +26,7 @@ Unless you explicitly state otherwise, any contribution intentionally submitted 
 
 [//]: # (badges)
 
-[crate-image]: https://img.shields.io/crates/v/dbl.svg
+[crate-image]: https://img.shields.io/crates/v/dbl.svg?logo=rust
 [crate-link]: https://crates.io/crates/dbl
 [docs-image]: https://docs.rs/dbl/badge.svg
 [docs-link]: https://docs.rs/dbl/
--- a/dbl/README.md
+++ b/dbl/README.md
@@ -26,7 +26,7 @@ Unless you explicitly state otherwise, any contribution intentionally submitted 
 
 [//]: # (badges)
 
-[crate-image]: https://img.shields.io/crates/v/dbl.svg?logo=rust
+[crate-image]: https://img.shields.io/crates/v/dbl.svg
 [crate-link]: https://crates.io/crates/dbl
 [docs-image]: https://docs.rs/dbl/badge.svg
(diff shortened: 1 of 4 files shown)


newpavlov (approved):
I have left some concerns [here](https://github.com/RustCrypto/utils/issues/1077#issuecomment-3152871718), but it's not a blocker and I am fine with merging it as-is.

tarcieri:
I think we should release it, but as noted in #1077 we can also continue working on a replacement in parallel, and it's easy to replace

newpavlov:
I think we can move forward with this after rebase (don't forget to update the release date and changelog in OP).

dbl/README.md
```diff
@@ -26,7 +26,7 @@ Unless you explicitly state otherwise, any contribution intentionally submitted
 
 [//]: # (badges)
 
-[crate-image]: https://img.shields.io/crates/v/dbl.svg
+[crate-image]: https://img.shields.io/crates/v/dbl.svg?logo=rust
```
newpavlov, line 29: BTW I am not sure about this change Rust log in a badge is bit too noisy for my personal taste and we probably should either do it for all crates or none.

tarcieri: I like it, it's still recognizable as the Rust logo and I've generally been adding it to new crates I release. But if you oppose it I can remove it.

newpavlov: It's better to remove it from this PR. If we are to change badges, It's better to do it in a dedicated PR for all crates in the repo.

PR #1205, opened 2025-08-25
cpufeatures: add compile time fallback for unsupported targets and target features

newpavlov:
The changes simplifies the code a bit and add a `cfg`-based fallback for all targets. This allows users to use the crate even if runtime detection was not implemented for a given target and target feature combination.

It ties the list of target features supported by the crate to `std`. This should not be a concern since we generally use target feature names defined by `std`.


newpavlov:
@heiher
I noticed that Loongarch features added in #955 differ from features supported by the [`is_loongarch_feature_detected!`](https://doc.rust-lang.org/std/arch/macro.is_loongarch_feature_detected.html) macro. Do you know why?

Additionally, `rustc --target=loongarch64-unknown-linux-gnu --print=target-features` prints the following list:
```
    crt-static           - Enables C Run-time Libraries to be statically linked.
    d                    - 'D' (Double-Precision Floating-Point).
    div32                - Assume div.w[u] and mod.w[u] can handle inputs that are not sign-extended.
    f                    - 'F' (Single-Precision Floating-Point).
    frecipe              - Support frecipe.{s/d} and frsqrte.{s/d} instructions..
    lam-bh               - Support amswap[_db].{b/h} and amadd[_db].{b/h} instructions..
    lamcas               - Support amcas[_db].{b/h/w/d}..
    lasx                 - 'LASX' (Loongson Advanced SIMD Extension).
    lbt                  - 'LBT' (Loongson Binary Translation Extension).
    ld-seq-sa            - Don't use load-load barrier (dbar 0x700)..
    lsx                  - 'LSX' (Loongson SIMD Extension).
    lvz                  - 'LVZ' (Loongson Virtualization Extension).
    relax                - Enable Linker relaxation.
    scq                  - Support sc.q instruction.
    ual                  - Allow memory accesses to be unaligned.
```
For example, there are no mentions of the `crypto` target feature.

I think we should change the module to follow `std`.

heiher:
@newpavlov You're right. The LoongArch target features in cpufeatures are indeed different from those used by rustc. In cpufeatures, they come from the Linux kernel's HWCAP set, but those don't map one-to-one with LLVM's compiler features. That's the root cause of the discrepancy. I agree we should align with std, and I'll update them accordingly.

**EDIT:**

#1206

newpavlov:
Why did you close this PR?

tarcieri:
Mistake, my bad

PR #1215, opened 2025-09-13
stack zeroization

### LINKED ISSUE #810: zeroize: stack zeroization support
It would be nice to be able to zeroize the stack, particularly being able to track how deep the stack grew while executing a particular section of code, and then zeroing out that whole portion of the stack.

This crate might provide the functionality needed to implement such functionality: https://crates.io/crates/psm

Additionally some proc macro annotation might be needed like `#[sensitive]` which could inject code to instrument the stack and track its depth.

nstilt1:
Uses `psm` to run a function/closure on a stack that is allocated on the heap, then zeroizes that separate stack when the execution finishes.

I attempted to zeroize the space between two pointers: one captured before the closure's call and one after, but upon zeroing the stack, there was a segmentation fault. I will have to remove that code, but it's there in the first commit if anyone thinks they can get it to run without a segmentation fault.

Closes #810 if this is up to standards.

### DIFF
--- a/zeroize_stack/src/heap/mmap.rs
+++ b/zeroize_stack/src/heap/mmap.rs
@@ -145,7 +145,7 @@ mod tests {
             let stack = ZeroizingHeapStack::new(kb);
             assert_eq!(
                 stack.stack_area().1, 
-                ((kb * 1024 + page_size() - 1) / page_size()) * page_size()
+                (kb * 1024).div_ceil(page_size()) * page_size()
             );
         }
     }


tarcieri:
I think it might be better if something like this started out as its own experimental crate, especially as `zeroize` is post-v1.0

nstilt1:
What should the crate be called? `stack_sanitizer`? And should this crate require `alloc` given that using the main function sort of requires the stack input to be on the heap? Should the call to `create_aligned_vec()` be included in `secure_crypto_call_heap()`? And `secure_crypto_call` isn't necessarily secure. Maybe it should be called `sanitize_crypto_call_stack`?

tarcieri:
> What should the crate be called?

`zeroize-stack` perhaps? Or `zeroize_stack` to match `zeroize_derive`.

Regarding the rest, spike it out however you'd like, but the difficult part will be the rationale for why it's sound.

nstilt1:
I can go with `zeroize_stack`. It fits. As for the rationale, the premise is exactly the same as another Rust crate, but it was only implemented for `x86_64`: [https://github.com/dsprenkels/eraser](https://github.com/dsprenkels/eraser).

With `psm`, we do not have to write the stack switching function in assemblies because it's already implemented for a number of architectures.

tarcieri:
Yeah, I've seen `eraser` in the past, it was mentioned in this paper: https://eprint.iacr.org/2023/1713.pdf

Interesting prototype of the idea.

newpavlov:
Does it pass Miri?

I left some ideas in the original issue.

nstilt1:
Darn. Miri wants me to use the following syntax due to how the `psm` crate is set up:

```rust
psm::psm_stack_manipulation! {
    yes {
        /// define fn with psm::on_stack
        pub unsafe fn blah(...) {...}
    }
    no {
        /// define fn without psm::on_stack, likely panic
        pub unsafe fn blah(...) { panic!("Stack manipulation not available on this platform")}
    }
}
```

After defining it like this, miri fails with the message "Stack manipulation not available on this platform". Likely because miri doesn't have a stack pointer or something... given that this crate was tied to the rust-lang account, I'd say their code is probably legit, and I'm not sure if Miri would be able to analyze all of the assembly without manually using different targets

nstilt1:
You can confirm that the test passes normally using `cargo test` but fails with `cargo miri test` due to the panic. Therefore this code is impossible to analyze with miri, and likely impossible to analyze with rudra given that it's edition 2024. But idek if rudra can analyze assembly files

nstilt1:
Further evidence for miri not working can be found in `psm`'s `build.rs`:

https://github.com/rust-lang/stacker/blob/7a3ff32d72bcd0a12a938abb21deddf9f1449cdc/psm/build.rs#L66

zeroize_stack/TODO.md
```diff
@@ -0,0 +1,47 @@
...
+            // Block on the future inside the heap stack
+            futures::executor::block_on(f())
+        })));
+    });
+
+    result.expect("Closure did not run")
+}
+```
+
+Copilot provided that code, but Gemini says that after the future is awaited, there will be no way for the program to know which stack to return to. Also, there is an open issue regarding async closures in `stacker` that has not been resolved after 7 months. https://github.com/rust-lang/stacker/issues/111
```
nstilt1, line 29: Do we agree with Gemini that this is either impossible or unsafe to perform? I can try to implement it and see if I get any segfaults, but async code is a low priority for me since basically all crypto operations are synchronous, and it can be ran in a way such that only crypto operations are done on separate stacks

newpavlov:
Personally, I do not plan to merge this PR in the near future. Instead I would like to play with the idea I mentioned in the issue.

nstilt1:
> Personally, I do not plan to merge this PR in the near future. Instead I would like to play with the idea I mentioned in the issue.

That's fine, but `asm!` is just outside of my scope. If you want to tackle safe stack bleaching, I would prefer to use your method over using the heap, but I just can't reliably prompt chat bots for `asm!` given that they frequently make things up, and since there are so many platforms that would need to be supported. If you have a good plan, then please go ahead and make a crate. I'll probably just use my solution until a better method is available.

tarcieri:
@nstilt1 as some general advice, if you want to work on this sort of concept, I'd suggest keeping it as minimal as possible for an MVP. No `async`. Just the simplest thing that works.

Also instead of asking LLMs, which don't actually know anything and can hallucinate answers, I think you'd get a lot more mileage out of actually asking the `psm` devs directly /cc @nagisa

nstilt1:
I found a bug in stacker. Kind of minor, but I revisited this crate with full support for wasm/hermitos and with guard pages for the other targets. Had to borrow some code from stacker because they seem to have worked out a lot of issues by using mmap instead of alloc-based allocations.

nstilt1:
I have an example for how to use the zeroizing heap stack with `switch_stacks()`, but it's a few hundred lines long... it's not even fully fleshed out, but it demonstrates how the stack can be reused, and how one could implement a full-on protocol using this. It would need to add `chacha20` as a dependency, along with `sha3` and `hkdf` using some of the new prerelease features for `chacha20`. But I think a few hundred lines might be too long... plus, I'm not sure whether the full example should go in the docs for `switch_stacks()` or in the module-level or crate-level documentation.

newpavlov:
Closing in favor of #1331

PR #1216, opened 2025-09-19
blobby: 0.4.0-pre.1

baloo:
This includes the binary file format changes included in #1207

This release is to remove the various `[patch.crates-io]` we have to support the new format.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,7 +4,7 @@ version = 4
 
 [[package]]
 name = "blobby"
-version = "0.4.0-pre.0"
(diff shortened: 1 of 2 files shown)


baloo:
cc @newpavlov I don't know if you anything else on the pipeline that might affect the file format.

newpavlov:
I wanted to finalize https://github.com/RustCrypto/traits/pull/1803 as part of https://github.com/RustCrypto/AEADs/issues/716 and then do the release, but I guess we can cut the release now.

PR #1217, opened 2025-09-20
block-padding: merge `Padding` and `RawPadding` traits

newpavlov:


newpavlov:
We need to also update the dependent downstream crates, but I think this change should not cause any problems.

@tarcieri 
Is there anything in `hybrid-array` which would help to remove the `unsafe` code used to transform `&[Array<u8, N>]` to `&[u8]`? It's the only `unsafe` code in the crate. The closes thing which I found is [`Array::as_flattened`](https://docs.rs/hybrid-array/latest/hybrid_array/struct.Array.html#method.as_flattened) (BTW note the typo in docs), but it transforms `&Array<Array<T, N>, M>`.

tarcieri:
@newpavlov we’d need to add something like `slice_as_flattened`. Also I don’t see the typo?

newpavlov:
Could you add it and cut new v0.4.1 release with it?

>I don’t see the typo

The docs contain `&Array<Array<T, N>, >>` while it should be `&Array<Array<T, N>, M>`

newpavlov:
I will merge this PR as-is for now and will open a separate issue for `unsafe` code removal.

PR #1218, opened 2025-09-20
block-buffer: simplify {Lazy,Eager}::split_blocks

kpp:
Use `split_at` so you don't need to calculate `tail` lenght in split_blocks fn. It reduces the amount of ptr arith, and there is no need to create `tail` from raw ptr, because it's created by `split_at` itself.

The change is no-panic:

```
use block_buffer::{EagerBuffer, Error, LazyBuffer, array::sizes::U8};
use no_panic::no_panic;

#[no_panic]
fn check_lazy() -> Result<(), Error> {
    let mut buffer = LazyBuffer::<U8>::try_new(&[0; 8])?;
    buffer.digest_blocks(&[0; 9], |_| {});
    Ok(())
}

#[no_panic]
fn check_eager() -> Result<(), Error> {
    let mut buffer = EagerBuffer::<U8>::try_new(&[0; 8])?;
    buffer.digest_blocks(&[0; 9], |_| {});
    Ok(())
}

fn main() {
    let _ = check_lazy();
    let _ = check_eager();
}
```

### DIFF
--- a/block-buffer/src/sealed.rs
+++ b/block-buffer/src/sealed.rs
@@ -62,15 +62,12 @@ impl Sealed for super::Eager {
     fn split_blocks<N: ArraySize>(data: &[u8]) -> (&[Array<u8, N>], &[u8]) {
         let nb = data.len() / N::USIZE;
         let blocks_len = nb * N::USIZE;
-        let tail_len = data.len() - blocks_len;
+
         // SAFETY: we guarantee that created slices do not point outside of `data`
         unsafe {
-            let blocks_ptr = data.as_ptr() as *const Array<u8, N>;
-            let tail_ptr = data.as_ptr().add(blocks_len);
-            (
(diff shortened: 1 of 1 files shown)


newpavlov:
This change results in a potential (IIUC unreachable) panic. Compare https://rust.godbolt.org/z/jMoWd1dM6 vs https://rust.godbolt.org/z/qd3MEajPj

kpp:
Fixed: https://rust.godbolt.org/z/jeMjTPMja. The new code generates no panic and it's shorter in instrs.

newpavlov:
Closing in favor of #1220. It does not use `unsafe` and compiles into the same number of instructions (though it uses conditional moves instead of branches): https://rust.godbolt.org/z/43xMT3ExP

kpp:
Thanks!

PR #1222, opened 2025-09-21
block-padding: remove unsafe code

### LINKED ISSUE #1221: block-padding: remove unsafe code
The only piece of `unsafe` code in the crate resides in the blanket implementation of the `Padding::unpad_blocks` method. It would be nice to remove it, but we need something like `slice_as_flattened` to be added into `hybrid-array` (i.e. equivalent of [`slice::as_flattened`](https://doc.rust-lang.org/std/primitive.slice.html#method.as_flattened)).

newpavlov:
Closes #1221

### DIFF
--- a/block-padding/src/lib.rs
+++ b/block-padding/src/lib.rs
@@ -8,6 +8,7 @@
     html_logo_url = "https://raw.githubusercontent.com/RustCrypto/media/6ee8e381/logo.svg",
     html_favicon_url = "https://raw.githubusercontent.com/RustCrypto/media/6ee8e381/logo.svg"
 )]
+#![deny(unsafe_code)]
 #![warn(missing_docs)]
 
 pub use hybrid_array as array;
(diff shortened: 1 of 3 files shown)


block-padding/src/lib.rs
```diff
@@ -64,7 +65,7 @@ pub trait Padding {
     /// Unpad data in the `blocks`.
     ///
     /// Returns `Err(UnpadError)` if the block contains malformed padding.
-    fn unpad_blocks<BlockSize: ArraySize>(
+    fn unpad_blocks<const N: usize, BlockSize: ArraySize<ArrayType<u8> = [u8; N]>>(
```
newpavlov, line 68: I should've seen it coming...

This bound is quite annoying ad will leak in the downstream crates, but I guess it should not be a problem in practice.

tarcieri: Yeah, these bounds are definitely annoying. They can be avoided at the cost of more unsafe code, which is something which has been done with other parts of `hybrid-array`

newpavlov: @tarcieri 
WDYT? I see 3 options:
1) Keep the old simple bound and `unsafe`
2) Tolerate the new bound
3) Try to relax the bounds in `hybrid-array` using `unsafe`

tarcieri:
@newpavlov not sure how you want to proceed, I opened #1224 which kind of duplicates this, but feel free to just update this PR and close #1224

PR #1225, opened 2025-09-23
block-padding: remove `PadType` and add `pad_detached` method

newpavlov:
`PadType` is used only in the default implementation of `unpad_blocks` and in `inout`. To handle the latter case `Padding::pad_detached` is added.


newpavlov:
`PadType` is used in `inout`, I will look into whether it's necessary.

PR #1242, opened 2025-10-28
[hex-literal]  Add new hex_custom_padding!() macro; enables custom padding between hex characters

mcronce:
Specifically, my use case is using key hashes from Android signing key reports, which are colon-delimited (e.g. `55:1A:14:14:BA:03:DA:1B:23:ED:EA:40:4A:BE:88:D7:C4:83:D1:E2:85:42:41:2C:22:73:D3:52:0B:6E:A7:5F`).  Removing the colons before use would work, but keeping them matching the signing key in source code provides a lot of value in terms of maintainability

Implemented by making the existing `hex!()` macro a wrapper around this, with no padding set

Probably needs docs; as long as this seems like something the community is comfortable with, I'll add a doc comment to the macro

### DIFF
--- a/hex-literal/src/lib.rs
+++ b/hex-literal/src/lib.rs
@@ -5,7 +5,7 @@
     html_favicon_url = "https://raw.githubusercontent.com/RustCrypto/media/6ee8e381/logo.svg"
 )]
 
-const fn next_hex_char(string: &[u8], mut pos: usize) -> Option<(u8, usize)> {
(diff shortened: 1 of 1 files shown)


newpavlov:
Hm. Maybe it would be better to just add `:` to the list of ignored characters? @tarcieri WDYT?

tarcieri:
Sure, seems fine

newpavlov:
Closing in favor of #1244.

mcronce:
Thanks!

PR #1260, opened 2025-12-20
ctutils: fix `clippy::bool_assert_comparison`

newpavlov:


newpavlov:
@tarcieri 
Note that the problematic test does not trigger our CI for some reason. We may need to add `--all-targets` for this.

newpavlov:
It was added in 1.53: https://rust-lang.github.io/rust-clippy/master/index.html#bool_assert_comparison But maybe it was not enabled by default?

PR #1264, opened 2025-12-20
ctutils: add `CtGt` and `CtLt`

tarcieri:
Adds traits for computing greater than/less than in constant time.

Uses ~~borrowing~~ overflowing subtraction as the implementation strategy for unsigned integers.

Also adds impls for `cmp::Ordering`.


tarcieri:
Ugh, MSRV is killing me here. Seems like `borrowing_sub` wasn't stabilized until 1.91.

Edit: aah, I want `overflowing_sub` anyway.

PR #1270, opened 2025-12-23
zeroize: fixup rust 1.92 warnings

baloo:
`#[derive(ZeroizeOnDrop)]` was throwing warnings like:
```
warning: value assigned to `curve` is never read
  --> src/crypto/ecdsa.rs:43:9
   |
43 |         curve: ECCCurve,
   |         ^^^^^
   |
   = help: maybe it is overwritten before being read?
   = note: `#[warn(unused_assignments)]` (part of `#[warn(unused)]`) on by de
```

This changes the tests to highlight the new warnings.

### DIFF
--- a/zeroize/tests/zeroize_derive.rs
+++ b/zeroize/tests/zeroize_derive.rs
@@ -125,7 +125,6 @@ impl Drop for ZeroizeNoDropEnum {
 
 #[test]
 #[cfg(feature = "alloc")]
-#[allow(unused_assignments)]
 fn derive_struct_skip() {
     #[derive(Zeroize, ZeroizeOnDrop)]
     struct Z {
@@ -156,7 +155,6 @@ fn derive_struct_skip() {
 
 #[test]
 #[cfg(feature = "alloc")]
-#[allow(unused_assignments)]
 fn derive_enum_skip() {
     #[derive(Zeroize, ZeroizeOnDrop)]
     enum Z {
--- a/zeroize_derive/src/lib.rs
+++ b/zeroize_derive/src/lib.rs
@@ -374,7 +374,7 @@ fn generate_fields(input: &DeriveInput, method: TokenStream) -> TokenStream {
         };
 
         quote! {
-            #[allow(unused_variables)]
+            #[allow(unused_variables, unused_assignments)]
             #binding => {
                 #(#method_field);*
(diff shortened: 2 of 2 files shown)


baloo:
If possible, a zeroize_derive 1.5.1 release with that PR would be appreciated!

tarcieri:
@baloo hmm, 1.5.0 is still unreleased, and includes an MSRV bump

We should probably get new releases of `zeroize` and `zeroize_derive` with that bump out, though I expect some pushback.

baloo:
Ha, right! I didn't even check.

Maybe I should backport this to 1.4 then.

baloo:
I don't really have a way to open a PR for a backport because there is no branch to target to.

I've made a commit:
 - https://github.com/RustCrypto/utils/compare/zeroize_derive-v1.4.2...baloo:push-mtmtvkvwzpzz?expand=1
 - https://github.com/baloo/utils/commit/ba99c6d28bf2ab5d10765d4a749d7292ecda7a08
 
If you could push a 1.4.3, that would be appreciated!

I've checked that commit on rpgp (and 1.92), and it doesn't report any warnings anymore.

dignifiedquire:
@tarcieri do you plan on doing a `1.4.x` release? I would like to avoid having to allow unused assignments in the whole crate

tarcieri:
Yeah, I'll take a look tomorrow

baloo:
a small release of zeroize itself (That bumps the dependency of zeroize_derive) would probably be nice to avoid issues with the minimal-versions checks as well.

tarcieri:
This should now be released as v1.4.3: https://github.com/RustCrypto/utils/pull/1310

@baloo re: pinning, I was planning on just cutting `zeroize` v1.9 and `zeroize_derive` v1.5 sometime early in January if that works

baloo:
Well, zeroize 1.9 would imply an MSRV bump. Which isn’t something necessarily doable for someone using zeroize 1.8.

this I why I was hinting at a zeroize 1.8.3 with a dependency on zeroize_derive 1.4.3.

otherwise this forces consumers to have a direct dependency on zeroize_derive to force 1.4.3 (to fix the minimal-versions checks). This is at least the case in rpgp, but there might be other consumers.

PR #1274, opened 2025-12-26
ctutils: `CtOption` methods for `crypto-bigint`

tarcieri:
Adds the methods needed to replace `ConstCtOption` in `crypto-bigint` with `ctutils::CtOption`:

- `some`
- `none`
- `into_option_copied`
- `filter_by`
- `as_inner_unchecked`
- `to_inner_unchecked`

Also adds a `Default` impl for `CtOption`.

Also adds `map!` and `unwrap_or!` macros to provide `const fn`-friendly combinator-like functionality.

### DIFF
--- a/ctutils/src/ct_option.rs
+++ b/ctutils/src/ct_option.rs
@@ -1,6 +1,36 @@
 use crate::{Choice, CtEq, CtSelect};
 use core::ops::{Deref, DerefMut};
 
+/// Helper macro for providing behavior like the [`CtOption::map`] combinator that works in
+/// `const fn` contexts.
+///
+/// Requires a provided `$mapper` function to convert from one type to another, e.g.
+///
+/// ```ignore
+/// const fn mapper(value: T) -> U
+/// ```
+#[macro_export]
+macro_rules! map {
+    ($opt:expr, $mapper:path) => {{ $crate::CtOption::new($mapper($opt.to_inner_unchecked()), $opt.is_some()) }};
+}
+
+/// Helper macro for providing behavior like the [`CtOption::unwrap_or`] combinator that works in
+/// `const fn` contexts.
+///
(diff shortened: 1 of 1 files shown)


ctutils/src/ct_option.rs
```diff
@@ -146,6 +195,13 @@ impl<T> CtOption<T> {
         optb
     }
 
+    /// Apply an additional [`Choice`] requirement to `is_some`.
+    #[inline]
+    pub const fn and_choice(mut self, is_some: Choice) -> Self {
+        self.is_some = self.is_some.and(is_some);
+        self
+    }
```
tarcieri: I'm wondering if there's a better name we can use for this. It looks a lot like `CtOption::filter` but instead of having a predicate callback, it just takes an option.

Maybe `filter_as`?

ctutils/src/ct_option.rs
```diff
@@ -164,6 +220,34 @@ impl<T> CtOption<T> {
...
+
+    /// DEPRECATED: legacy compatibility method for `crypto-bigint`
+    #[inline]
+    #[deprecated(since = "0.1.1", note = "use `as_inner_unchecked` instead")]
+    pub const fn components_ref(&self) -> (&T, Choice) {
+        // Since Rust is not smart enough to tell that we would be moving the value,
+        // and hence no destructors will be called, we have to return a reference instead.
+        // See https://github.com/rust-lang/rust/issues/66753
+        (&self.value, self.is_some)
+    }
```
tarcieri: Ideally I can get rid of this before merging, replacing it with `as_inner_unchecked`

ctutils/src/ct_option.rs
tarcieri: Went with `filter_by`

PR #1291, opened 2025-12-27
inout: require `block-padding` v0.4.2

tarcieri:
The other versions are yanked, but nothing is forcing an upgrade if people already have older versions in their Cargo.lock.

The problematic crates are ones like `cipher` that are using it vicariously, so we need to pin it here to force an upgrade.


tarcieri:
Ran into this upgrading https://github.com/RustCrypto/block-modes and didn't want others to have to deal with it

PR #1295, opened 2025-12-28
cmov: add proptests

tarcieri:
Tests the implementation against random cases to ensure correctness


tarcieri:
That's no good. Seems like `cmoveq`/`cmovne` are broken on PPC32?

PR #1297, opened 2025-12-28
cmov: test with miri on `powerpc-unknown-linux-gnu`

tarcieri:
Trying to catch the same bug from #1295 with the existing test suite


tarcieri:
Well, it passes, but didn't catch the bug

Issue #1298, opened 2025-12-28
cmov: `cmoveq`/`cmovne` proptest failure on `powerpc-unknown-linux-gnu`


tarcieri:
Possible big endian issue in the portable implementation? Discovered via https://github.com/RustCrypto/utils/pull/1295

The tests look like this, so only `cmoveq`/`cmovne` are impacted, whereas `cmovz`/`cmovnz` seem to be OK:

```console
test i128::cmoveq_works ... ok
test i128::cmovne_works ... ok
test i128::cmovnz_works ... ok
test i128::cmovz_works ... ok
test i16::cmoveq_works ... ok
test i16::cmovne_works ... FAILED
test i16::cmovnz_works ... ok
test i16::cmovz_works ... ok
test i32::cmoveq_works ... FAILED
test i32::cmovne_works ... FAILED
test i32::cmovnz_works ... ok
test i32::cmovz_works ... ok
test i64::cmoveq_works ... FAILED
test i64::cmovne_works ... FAILED
test i64::cmovnz_works ... ok
test i64::cmovz_works ... ok
test i8::cmoveq_works ... ok
test i8::cmovne_works ... ok
test i8::cmovnz_works ... ok
test i8::cmovz_works ... ok
test u128::cmoveq_works ... ok
test u128::cmovne_works ... ok
test u128::cmovnz_works ... ok
test u128::cmovz_works ... ok
test u16::cmoveq_works ... FAILED
test u16::cmovne_works ... FAILED
test u16::cmovnz_works ... ok
test u16::cmovz_works ... ok
test u32::cmoveq_works ... FAILED
test u32::cmovne_works ... FAILED
test u32::cmovnz_works ... ok
test u32::cmovz_works ... ok
test u64::cmoveq_works ... FAILED
test u64::cmovne_works ... FAILED
test u64::cmovnz_works ... ok
test u64::cmovz_works ... ok
test u8::cmoveq_works ... ok
test u8::cmovne_works ... ok
test u8::cmovnz_works ... ok
test u8::cmovz_works ... ok
```

Here are the failing cases:

From https://github.com/RustCrypto/utils/actions/runs/20548544186/job/59022859407?pr=1295 while it lasts

```console
---- i16::cmovne_works stdout ----
proptest: FileFailurePersistence::SourceParallel set, but failed to find lib.rs or main.rs
proptest: FileFailurePersistence::SourceParallel set, but failed to find lib.rs or main.rs
[...]

PR #1299, opened 2025-12-28
 cmov: fix `aarch64` ASM bug; add regression test

tarcieri:
The implementation was using the wrong register type when invoking `csel` for `u64` (`:w` which is 32-bit, should use `:x` for 64-bit)

This captures the bug in a regression test and fixes it.

### DIFF
--- a/cmov/src/aarch64.rs
+++ b/cmov/src/aarch64.rs
@@ -1,8 +1,9 @@
 use crate::{Cmov, CmovEq, Condition};
 use core::arch::asm;
 
+/// Conditional select
 macro_rules! csel {
-    ($csel:expr, $dst:expr, $src:expr, $condition:expr) => {
+    ($cmp:expr, $csel:expr, $dst:expr, $src:expr, $condition:expr) => {
         unsafe {
             asm! {
                 "cmp {0:w}, 0",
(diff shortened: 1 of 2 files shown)


tarcieri:
The PPC failure is probably the same issue as #1298.

GitHub Actions killed the `aarch64` runs but they're also failing:

https://github.com/RustCrypto/utils/actions/runs/20548849075/job/59023609534?pr=1299

```console
Run cross test --target aarch64-unknown-linux-gnu
[...]
     Running tests/regression.rs (/target/aarch64-unknown-linux-gnu/debug/deps/regression-d4120ddf24ad0571)

running 1 test
test u64_cmoveq ... FAILED

failures:

---- u64_cmoveq stdout ----

thread 'u64_cmoveq' panicked at cmov/tests/regression.rs:12:5:
assertion `left == right` failed
  left: 1
 right: 0
```

I sprinkled some `dbg!` into the `u64` impls in `aarch64.rs` like so:

```rust
#[inline]
fn cmoveq(&self, rhs: &Self, input: Condition, output: &mut Condition) {
    dbg!(self, rhs, input, &output);
    csel_eq!("csel {3:w}, {4:w}, {5:w}, EQ", self, rhs, input, output);
}
```

It's definitely getting called, here's the output:

```console
---- u64_cmoveq stdout ----
[cmov/src/aarch64.rs:110:9] self = 9367487224930631680
[cmov/src/aarch64.rs:110:9] rhs = 0
[cmov/src/aarch64.rs:110:9] input = 1
[cmov/src/aarch64.rs:110:9] &output = 0

thread 'u64_cmoveq' panicked at cmov/tests/regression.rs:12:5:
assertion `left == right` failed
  left: 1
 right: 0
```

cc @brxken128 @codahale who worked on the `aarch64` impl

tarcieri:
It seems the registers should be `:x` instead of `:w` but that doesn't seem to actually fix the problem

Edit: aah, looks like the other macros hardcode `:w` too and that might need to be changed to `:x`?

Edit again: confirm changing all of the asm registers from `:w` to `:x` fixes the problem

PR #1319, opened 2025-12-30
CI: actually test `zeroize_derive`

tarcieri:
It seems there wasn't actually any CI for `zeroize_derive` after it was extracted from `zeroize/derive` into a toplevel `zeroize_derive` directory, other than the vicarious tests that are in `zeroize` itself.

This runs a separate job just for the custom derive tests.

NOTE: they presently seem to be failing!


tarcieri:
Well, that at least captured the test failure in CI: https://github.com/RustCrypto/utils/actions/runs/20605171959/job/59179321072?pr=1319

PR #1331, opened 2026-01-07
zeroize: add `zeroize_stack` function

### LINKED ISSUE #810: zeroize: stack zeroization support

newpavlov:
Adds a simple `optimization_barrier`-based function for stack zeroization. It requires an explicitly provided size since erasing the whole stack (2-8 MiB) can be expensive and it would be an overkill in most cases. Additionally, detection of remaining stack size (e.g. using `pthread_attr_getstack`) is platform-dependent and not available on all targets.

Closes #810

### DIFF
--- a/zeroize/CHANGELOG.md
+++ b/zeroize/CHANGELOG.md
@@ -7,12 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## 1.9.0 (unreleased)
 ### Added
 - `optimization_barrier` function ([#1261])
+- `zeroize_stack` function ([#1331])
 
 ### Changed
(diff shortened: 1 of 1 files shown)


newpavlov:
Here you can see the function in action: https://rust.godbolt.org/z/Yd9hce7Yq

tarcieri (approved):

PR #1332, opened 2026-01-14
cmov: fix portable `cmovnz` for `thumbv6m-none-eabi`

NicsTr:
`thumbv6m-none-eabi` (Cortex M0, M0+ and M1) compiler emits non-constant time assembly when using `cmovnz` (portable version).

This commit fixes it.

### DIFF
--- a/.typos.toml
+++ b/.typos.toml
@@ -2,3 +2,7 @@
 extend-exclude = [
     ".git/"
 ]
+
+[default.extend-words]
+# Prefix of Conditional MOVe
+"CMO" = "CMO"


tarcieri:
Weird, that's an annoying thing to have typos fail on.

@NicsTr can you edit `.typos.toml` and add something like this?

https://github.com/RustCrypto/traits/blob/master/.typos.toml#L14-L16

newpavlov:
`thumbv6m` has stable `asm!`, no?

tarcieri:
Yeah, I guess so. Perhaps we can use `asm!` instead of `black_box` on a subset of platforms which have stable `asm!` but don't have a backend written with architecture-specific predication instructions

PR #1333, opened 2026-01-14
cmov v0.4.4

tarcieri:
### Security
- Fix non-constant-time assembly being emitted from portable backend on `thumbv6m-none-eabi` ([#1332])

[#1332]: https://github.com/RustCrypto/utils/pull/1332


tarcieri:
cc @NicsTr

PR #1334, opened 2026-01-14
cmov: use `asm!` barrier in portable backend (where stable)

tarcieri:
On targets where `asm!` is stable but we don't have a backend using architecture-specific predication instructions, uses `asm!` to insert an fence/barrier which should hopefully accomplish the intended goal as well or better than `core::hint::black_box`, using a similar approach to what was introduced in `zeroize` in #1252.

Suggested by @newpavlov in #1332.

### DIFF
--- a/cmov/src/portable.rs
+++ b/cmov/src/portable.rs
@@ -7,6 +7,43 @@
 // TODO(tarcieri): more optimized implementation for small integers
 
 use crate::{Cmov, CmovEq, Condition};
+
+// Use `asm!` on architectures where it's stable but we don't have a custom-written backend
+#[cfg(all(
+    not(miri),
+    any(
+        target_arch = "arm",
+        target_arch = "arm64ec",
+        target_arch = "loongarch64",
+        target_arch = "riscv32",
+        target_arch = "riscv64",
+        target_arch = "s390x"
+    )
+))]
+fn black_box<T: Copy>(val: T) -> T {
+    #[allow(trivial_casts)]
+    unsafe {
+        core::arch::asm!(
+            "# {}",
+            in(reg) &val as *const T as *const (),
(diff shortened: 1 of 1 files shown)


tarcieri:
@NicsTr can you verify this still accomplishes the intended codegen?

NicsTr:
When testing `cmovnz` with `thumbv6m-none-eabi`, it introduces a branch:

```asm
.section .text.not_ct::test_ct_cmov,"ax",%progbits
	.globl	not_ct::test_ct_cmov
	.p2align	1
	.type	not_ct::test_ct_cmov,%function
	.code	16
	.thumb_func
not_ct::test_ct_cmov:
	.fnstart
	.cfi_sections .debug_frame
	.cfi_startproc
	.save	{r4, r5, r7, lr}
	push {r4, r5, r7, lr}
	.cfi_def_cfa_offset 16
	.cfi_offset lr, -4
	.cfi_offset r7, -8
	.cfi_offset r5, -12
	.cfi_offset r4, -16
	.setfp	r7, sp, #8
	add r7, sp, #8
	.cfi_def_cfa r7, 8
	.pad	#4
	sub sp, #4
	uxtb r3, r2
	subs r2, r3, #1
	mov r4, r3
	sbcs r4, r2
	ldrb r2, [r0]
	mov r5, sp
	strb r4, [r5]
	@APP
	@ r5
	@NO_APP
	cmp r3, #0
	bne .LBB0_2
	mov r1, r2
.LBB0_2:
	strb r1, [r0]
	add sp, #4
	pop {r4, r5, r7, pc}
```

Is this approach of `black_box` (using inline `asm`), supposed to be more robust than the `core::hint::black_box`?

tarcieri:
That was the hypothesis but perhaps it isn't true

PR #1335, opened 2026-01-15
cmov: extract `maskgen32` and `maskgen64` in portable backend

tarcieri:
Extracts some functions in the portable backend that can both be easily tested and easily substituted with some platform-specific idealized assembly for emulating CMOVNZ that everything else can build on top of.

Also adds tests in CI for an `armv7-unknown-linux-gnueabi` target we can use for verifying correct behavior on 32-bit platforms.

### DIFF
--- a/cmov/src/portable.rs
+++ b/cmov/src/portable.rs
@@ -7,27 +7,26 @@
 // TODO(tarcieri): more optimized implementation for small integers
 
 use crate::{Cmov, CmovEq, Condition};
-use core::hint::black_box;
 
 /// Bitwise non-zero: returns `1` if `x != 0`, and otherwise returns `0`.
 macro_rules! bitnz {
     ($value:expr, $bits:expr) => {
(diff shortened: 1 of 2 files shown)


cmov/src/portable.rs
```diff
@@ -99,3 +98,40 @@ impl CmovEq for u64 {
...
+
+/// Return a [`u32::MAX`] mask if `condition` is non-zero, otherwise return zero for a zero input.
+pub fn nzmask32(condition: Condition) -> u32 {
+    bitnz!(condition as u32, u32::BITS).wrapping_neg()
+}
+
+/// Return a [`u64::MAX`] mask if `condition` is non-zero, otherwise return zero for a zero input.
+pub fn nzmask64(condition: Condition) -> u64 {
+    bitnz!(condition as u64, u64::BITS).wrapping_neg()
+}
```
tarcieri: Note: these actually return the opposite of what we were doing for: `wrapping_neg()` vs `wrapping_sub(1)`

I think this approach is a little more straightforward.

PR #1338, opened 2026-01-15
cmov: add `testnz!` macro in the portable backend

tarcieri:
Instead of having the `bitnz!` macro which inserts an expression, this instead replaces it with a `testnz!` macro and uses it to write `testnz32` and `testnz64` functions, then changes the places the macro was being invoked to call the functions instead.

This reduces our dependence on the macro to just expressing boilerplate functions that can't be written generically (e.g. there's no trait for `wrapping_neg`)

### DIFF
--- a/cmov/src/portable.rs
+++ b/cmov/src/portable.rs
@@ -8,13 +8,6 @@
 
 use crate::{Cmov, CmovEq, Condition};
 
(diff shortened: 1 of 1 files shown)


tarcieri:
Actually, let me try this again with no macros whatsoever

PR #1336, opened 2026-01-15
cmov: add `asm!` optimized `masknz32` for ARM32

tarcieri:
In #1332 we ran into LLVM inserting branches in this routine for `thumbv6m-none-eabi` targets. It was "fixed" by fiddling around with `black_box` but that seems brittle.

In #1334 we attempted a simple portable `asm!` optimization barrier approach but it did not work as expected.

This instead opts to implement one of the fiddliest bits, mask generation, using ARM assembly instead. The resulting assembly is actually more efficient than what rustc/LLVM outputs and avoids touching the stack pointer.

It's a simple enough function to implement in assembly on other platforms with stable `asm!` too, but this is a start.

PR #1347, opened 2026-01-15
cmov v0.4.5

tarcieri:
### Changed
- Introduce small ARM32 `asm!` optimization which also guarantees constant-time operation ([#1336], [#1346])

[#1336]: https://github.com/RustCrypto/utils/pull/1336
[#1346]: https://github.com/RustCrypto/utils/pull/1346


tarcieri:
FYI @NicsTr, it would be great if you could double-check the ARM32 codegen

NicsTr:
I can confirm that the generated code for `thumbv6m-none-eabi` is the one expected!

PR #1452, opened 2026-02-06
Remove `opaque-debug` and `collectable`

newpavlov:
`opaque-debug` is no longer used in RustCrypto crates and its uses were replaced by explicit `fmt::Debug` implementations, so the crate is de facto deprecated. I intend to release empty `opaque-debug` v0.4.0 with deprecation notice.

`collectable` is also abandoned.


tarcieri (approved):

tarcieri:
Could probably remove `collectable` too as an abandoned experiment

srfsh:
i just discovered collectable. i think it was meant to bridge the gap between `heapless` and `alloc` variants of data structures like `Vec`, `String`, no? if that's the case, why is it removed? i was wrapping my head around this idea and just realized it existed. why did the experiment fail? the idea seems brilliant to me, letting the library authors using it for being agnostic over `heapless` and `alloc` while the library consumers decide what they would like to use

tarcieri:
The main functionality `collectable` was intended for was as a more general replacement for the `aead::Buffer` trait, but it was never really "completed" or maintained, and the `aead::Buffer` trait was good enough for our purposes.

PR #1470, opened 2026-04-09
Fix output for 16-bit targets

kayabaNerve:
`tokens32` being used for both 16 and 32 bits appears to be an artifact from copying the prior block and modifying it (as reasonable for such development), but itself unsound.

---

I used the web UI for this, just to fire this off quickly as an exceptionally minor change, but that does mean it's completely untested.

### DIFF
--- a/cpubits/src/lib.rs
+++ b/cpubits/src/lib.rs
@@ -259,7 +259,7 @@ macro_rules! cpubits {
                 // WASM
                 target_arch = "wasm32",
             ))]
-            16 => { $( $tokens32 )* }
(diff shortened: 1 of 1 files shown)


kayabaNerve:
CI passing is always a good sign :tada: 

I caught this while beginning working on a `cargo vet` of my entire tree, a task of such size it may never be finished. I hope for a `cpubits 0.1.1` soon so I can give it a clean vet though :)

tarcieri (approved):
Thanks! Unfortunately I wasn’t able to figure out a good solution for 16-bit CI

PR #1481, opened 2026-05-15
bitref: bit-level reference types

tarcieri:
Adds a crate providing a `&BitSlice`/`&mut BitSlice` type which is constructable from `&[u8]` but provides slicing at the granularity of individual bits. The name of the crate is a play on `bitvec`, which provides a similar type. However, the implementation in this crate is significantly simpler with a much smaller code surface and minimal use of `unsafe` code.

The implementation is a generalization of RustCrypto/formats#2300 which sought to implement a similar data structure as a reference type for representing ASN.1 BIT STRINGs. However, using this approach was deferred because the implementation relies on a conversion which is sound under Tree Borrows (as verified by Miri) but unsound under Stacked Borrows as it loses provenance. See rust-lang/unsafe-code-guidelines#134

There are several places such a data structure is potentially useful for RustCrypto projects. Beyond the previously mentioned ASN.1 BIT STRING use case, being able to iterate over bits is useful in many numerical algorithms with applications in cryptography, notably in `crypto-bigint` and for elliptic curves.

Elliptic curve scalar multiplication is generally implemented as a loop over the bits of a scalar. Having an iterator type for this purpose avoids problems relating to the endianness of how scalars are serialized when implementing generic scalar multiplication algorithms, e.g. wNAF (see RustCrypto/group#12).

Given the current open soundness story, I'm not rushing to use this in `crypto-bigint` until that changes. Where we could use it today though is as an optional dependency to `der`, where it can act as an ASN.1 BIT STRING type, but implement `ToOwned` producing a `der::asn1::BitString` (which, to make `ToOwned` work, needs to impl `Borrow<BitSlice>`).

This would make it optionally possible to use `Cow` for copy-on-write BIT STRINGs today with `BitSlice` as the borrowed form, but leaving the preferred default data structure for that purpose as `der::asn1::BitStringRef`, which is a lifetime-parameterized struct that avoids the open soundness questions around `BitSlice`.

From there we can see what develops around the soundness story and SB/TB discrepancy, and beyond that new Rust features like custom DSTs which may make expressing structures like this less of a hack.

### DIFF
--- a/.github/workflows/bitref.yml
+++ b/.github/workflows/bitref.yml
@@ -1,0 +1,135 @@
+name: bitref
+
+on:
+  pull_request:
+    paths:
+      - ".github/workflows/bitref.yml"
+      - "bitref/**"
+      - "Cargo.*"
+  push:
+    branches: master
+
+permissions:
+  contents: read
+
+defaults:
+  run:
+    working-directory: bitref
+
+env:
+  CARGO_INCREMENTAL: 0
+  RUSTFLAGS: "-Dwarnings"
+
+# Cancels CI jobs when new commits are pushed to a PR branch
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  build-no-std:
+    strategy:
+      matrix:
+        target:
+          - riscv32i-unknown-none-elf
+          - thumbv7em-none-eabi
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+      - uses: RustCrypto/actions/cargo-cache@master
+      - uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: stable
+          targets: ${{ matrix.target }}
+      - uses: RustCrypto/actions/cargo-hack-install@master
+      - run: cargo build --target ${{ matrix.target }}
+
+  minimal-versions:
+    uses: RustCrypto/actions/.github/workflows/minimal-versions.yml@master
+    with:
+      working-directory: ${{ github.workflow }}
+
+  test:
+    strategy:
+      matrix:
+        include:
+          - rust: 1.85.0 # MSRV
+          - rust: stable
+    runs-on: "ubuntu-latest"
+    steps:
+      - uses: actions/checkout@v6
+      - uses: RustCrypto/actions/cargo-cache@master
+      - uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: ${{ matrix.rust }}
+      - run: cargo test
+
+  # Test using `cargo careful`
+  test-careful:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+      - uses: dtolnay/rust-toolchain@nightly
+      - run: cargo install cargo-careful
+      - run: cargo careful test
+
+  # Cross-compiled tests
+  test-cross:
+    strategy:
+      matrix:
+        include:
+          # ARM32
+          - target: armv7-unknown-linux-gnueabi
+            rust: 1.85.0 # MSRV
+          - target: armv7-unknown-linux-gnueabi
+            rust: stable # MSRV
+          # PPC32
+          - target: powerpc-unknown-linux-gnu
+            rust: 1.85.0 # MSRV
+          - target: powerpc-unknown-linux-gnu
+            rust: stable
+          # RISCV64
+          - target: riscv64gc-unknown-linux-gnu
+            rust: stable
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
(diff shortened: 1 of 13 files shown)


thomcc:
It seems dubious to use techniques which don't pass SB in a foundational crate like this. That means none of the users of this (perhaps transitively) can use SB. Given that SB is safer than TB (in that there are things allowed in TB which are likely to be UB eventually, which is not really true for SB), that doesn't seem good.

You also don't need the `UnsafeCell<()>`. Just `()` should be fine, since you don't support a `&[Cell<u8>]`-backed BitSlice with interior mutability (`bitvec` does, but the difference between it and the normal BitSlice was tracked in PhantomData types alone, so `rustc`/LLVM couldn't tell, and just saw that there were writes to a pointer which came from an immutable slice.

tarcieri:
@thomcc this is not nor are there plans for it to be a "foundational crate" until the soundness story is resolved.

As the commit message notes, I am not rushing to use this, and the `der` crate already provides a `BitStringRef<'_>` type with equivalent functionality but the inability to use reference patterns with it as it's a lifetime-parameterized struct, and likewise the PR to actually convert `BitStringRef` to this form was closed explicitly because the soundness story isn't there yet: https://github.com/RustCrypto/formats/pull/2300

tarcieri:
To expand on that, the main reasons I want to put this under RustCrypto:

1. There are already people using `bitvec` for cryptography (though [we aren't anymore](https://github.com/RustCrypto/traits/pull/2417)) and this at least offers a less-bad partial replacement which passes under Miri with TB if not SB
2. The `der` use case could be useful today as an optional, caveat-adorned alternative to [`der::asn1::BitStringRef<'a>`](https://docs.rs/der/latest/der/asn1/struct.BitStringRef.html)
3. I would aspirationally really like to have something like this for `crypto-bigint`, and it illustrates a "here's something I'd like to use for cryptography but can't because the soundness story isn't quite there" use case when it comes to nailing down the opsem

I am happy to add an additional note about this all in the README.md, saying we aren't using it anywhere yet because of the open soundness issues.

But if there are continued objections, I can just locate it under @tarcieri rather than @RustCrypto. But I thought I'd try here first.

newpavlov:
I can not say I am familiar with this area, but is there a reason to fixate on working with the bit slice types via references?

Can't we introduce quasi-reference types like this instead?

```rust
// raw pointers are used for the mut variant
struct BitSlice(Mut)<'a> {
    origin: *const/mut u8,
    bit_offset: usize,
    bit_len: usize,
    _pd: PhantomData<&'a ()>,
}
```

By making `BitSliceMut` non-`Send`able it should be possible to construct a safe API to deal with the referenced data. I am not sure how it would interact with `ToOwned` and CoW, but it may be possible to introduce an owned variant as well. In other words, it would be an evolution of the existing `BitStringRef`.

tarcieri:
@newpavlov as I [noted yesterday](https://github.com/RustCrypto/utils/pull/1481#issuecomment-4462563661) that's how the existing `der::asn1::BitStringRef<'a>` type works, however it cannot impl patterns like `Borrow`/`ToOwned` because it's not a reference type. There are other traits which necessitate borrows noted in the README which this unlocks.

So, what you're describing already exists. This is an alternative which can be plugged into something like [`SubjectPublicKeyInfo`](https://docs.rs/spki/latest/spki/struct.SubjectPublicKeyInfo.html) generically (provided it impl'd the required traits) which could work with `Cow` and provide copy-on-write support.

There are other ergonomic niceties, like not having to have a `BitSlice`/`BitSliceMut` split. You can't paper over that with `Deref` either, because in such a case `BitSlice` won't be a reference.

newpavlov:
>however it cannot impl patterns like Borrow/ToOwned because it's not a reference type

But do we _have_ to use them? IIUC your main motivation is to be able to use `std::borrow::Cow`, but I think we can easily introduce something like `BitCow` and some associated traits to work around it. IMO dealing with small ergonomic complications is better than relying on code with dubious soundness.

I don't have high hopes for the language improving in this area in the following years, so if your goal is to use this crate only for [illustration purposes](https://github.com/RustCrypto/utils/pull/1481#issuecomment-4462670731), I don't think it makes sense to keep it in RustCrypto/utils.

>like not having to have a BitSlice/BitSliceMut split.

I think there were some proposals for improving this and general convenience of working with custom reference-like types. Though I can not name them out of my head.

tarcieri:
> But do we have to use them? IIUC your main motivation is to be able to use std::borrow::Cow, but I think we can easily introduce something like BitCow and some associated traits to work around it

That's the [solution already proposed in RustCrypto/formats#2298](https://github.com/RustCrypto/formats/issues/2298#issuecomment-4313995202), but it's not a particularly good one. It somewhat annoyingly has to duplicate core functionality with incompatible types/traits.

Anyway, it seems like there's enough opposition to having this in @RustCrypto I'll go ahead and close this and publish it under @tarcieri instead.

PR #1486, opened 2026-06-05
cpufeatures: gracefully degrade on targets without a detection backend

pkubaj:
cpufeatures emitted a hard `compile_error!` on any target other than
aarch64/loongarch64/x86/x86_64. This breaks otherwise-portable crates that
depend on cpufeatures unconditionally but only invoke `cpufeatures::new!`
under arch-gated cfgs (e.g. `#[cfg(target_arch = "x86_64")]`), so the
detection macros are never actually expanded on other architectures
(powerpc, powerpc64, etc.) — the crate simply fails to compile.

Add a `fallback` module providing the `__unless_target_features!` and
`__detect_target_features!` macros for those targets. Runtime detection is
unavailable there, so feature checks rely solely on compile-time target
features and otherwise report features as absent — mirroring the existing
"other OS" fallback already used within the aarch64 backend.

### DIFF
--- a/cpufeatures/src/fallback.rs
+++ b/cpufeatures/src/fallback.rs
@@ -1,0 +1,30 @@
+//! Fallback for targets without a runtime CPU feature detection backend.
+//!
+//! On architectures cpufeatures has no detection support for (e.g.
+//! `powerpc`/`powerpc64`), runtime detection is unavailable, so feature checks
+//! fall back to compile-time target features only.
+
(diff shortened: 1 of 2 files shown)


newpavlov:
You arguably should not use `cpufeatures` on unsupported targets, but I understand why having a compile-time fallback could be useful. I tried to do a similar thing in #1205, but I forgot to finish it.

PR #1517, opened 2026-08-13
digest-io: add support for AsyncRead/AsyncWrite

baloo:


newpavlov:
Personally, I would prefer to not have async support in `digest-io`, at least as long as async traits are not part of `std`. A separate crate could be a better place for it.

PR #1518, opened 2026-08-15
digest-io-async: initial commit

baloo:
This introduces a new crate providing `HashReader`/`HashWriter` wrappers implementing `AsyncRead`/`AsyncWrite`, mirroring `digest-io`'s sync API.

This only implement the tokio's async API for now.

The intent is to fold it in `digest-io` when the
`AsyncReader`/`AsyncWriter` eventually make their way into `std`.


baloo:
This is a follow up on #1517

digest-io-async/LICENSE-MIT
```diff
@@ -0,0 +1,25 @@
+Copyright (c) 2025 The RustCrypto Project Developers
```
newpavlov: Change the year to 2026.

digest-io-async/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+
+[dev-dependencies]
+bytes = "1"
+futures = "0.3"
+sha2 = "0.11"
+tokio = { version = "1", features = ["macros", "rt"] }
+tokio-util = { version = "0.7", features = ["io"] }
+
+[features]
+tokio = ["dep:tokio", "dep:pin-project-lite"]
```
newpavlov: I think we can remove the `tokio` crate feature for now. If we are to add support for other executors we could bring it back in a future breaking release.

newpavlov (commented):
This looks as a fine addition to me. I will leave the final approval to @tarcieri since I (thankfully) barely deal with async Rust code.

Cargo.toml
```diff
@@ -9,6 +9,8 @@ members = [
     "ctutils",
     "cpufeatures",
     "dbl",
+    "digest-io",
```
newpavlov: `digest-io` was intentionally excluded from the workspace (see below). Do the same for `digest-io-async`.

nazar-pc (commented):
> This only implement the tokio's async API for now, and it is gated behind a `tokio` feature.

Right now `tokio` is a mandatory dependency and there are no feature gates anywhere

tarcieri (approved):

newpavlov:
GitHub experiences issues right now, so I will postpone merge until they are resolved.

PR #1528, opened 2026-09-02
cpufeatures: fix x86 CPUID and XSAVE state checks

### LINKED ISSUE #1510: x86: guard CPUID leaf 7 by the maximum basic leaf; AVX detection may need XMM+YMM XCR0 state
## Versions inspected

- `cpufeatures` 0.2.17
  - crates.io SHA-256: `59ed5838eebb26a2bb2e58f6d5b5316989ae9d08bab10e0e6d103e656d1b0280`
  - upstream commit recorded by the package: `9d92d5e95ab4c07c5d8bfd024bf2a17e96d20feb`
- `cpufeatures` 0.3.0
  - crates.io SHA-256: `8b2a41393f66f16b0823bb79094d54ac5fbd34ab292ddafb9a0456ac9f87d201`
  - upstream commit recorded by the package: `31e6ec3727f11f9f5e0c106e282014653a4d7bb7`

While reviewing the x86 feature-detection implementation, I found two independent questions. I am reporting them together for context, but I would be happy to split them into separate issues if that is easier to review.

### Summary

Both inspected versions query structured feature leaf 7 directly, without first checking the maximum basic CPUID leaf reported by `CPUID(0).EAX`.

This appears not to fail closed on processors or virtual CPUs whose maximum basic leaf is below 7.

Updating from 0.2.17 to 0.3.0 does not currently address this: 0.3.0 retains the direct leaf-7 query and additionally queries subleaf 7.1.

### Source evidence

In 0.2.17, `cpufeatures/src/x86.rs:47-58` constructs:

```rust
[cpuid(1), cpuid_count(7, 0)]
```

Source:

https://github.com/RustCrypto/utils/blob/9d92d5e95ab4c07c5d8bfd024bf2a17e96d20feb/cpufeatures/src/x86.rs#L47-L58

In 0.3.0, `cpufeatures/src/x86.rs:41-51` constructs:

```rust
[cpuid(1), cpuid_count(7, 0), cpuid_count(7, 1)]
```

Source:

https://github.com/RustCrypto/utils/blob/31e6ec3727f11f9f5e0c106e282014653a4d7bb7/cpufeatures/src/x86.rs#L41-L51

A package-wide search of the Rust sources in both releases found no call to `__get_cpuid_max`, no `cpuid(0)`, and no equivalent maximum-basic-leaf guard.

The Rust `core::arch::__cpuid_count` documentation states that when the requested leaf is above the maximum supported basic or extended leaf, the result is the information for the highest supported basic information leaf, using the supplied subleaf:

https://doc.rust-lang.org/core/arch/x86_64/fn.__cpuid_count.html

Therefore an unsupported query for leaf 7 is not guaranteed to produce zero feature bits. Bits belonging to the highest supported basic leaf could instead be interpreted using the leaf-7 layout.

### Observed behavior

The detector always obtains leaf 7 before evaluating the requested feature predicates. If the reported maximum basic leaf is less than 7, the returned registers may describe a different leaf, but the `check!` rules interpret them as leaf-7 feature flags.

### Expected behavior

Unsupported structured-feature leaves and subleaves should be represented as zero / unavailable before their bits are interpreted.

Conceptually:

```text
max_basic_leaf = CPUID(0).EAX

if max_basic_leaf >= 7:
    leaf7_0 = CPUID(7, 0)
else:
    leaf7_0 = zero / features unavailable

if max_basic_leaf >= 7 and leaf7_0 reports subleaf 1 as supported:
    leaf7_1 = CPUID(7, 1)
else:
    leaf7_1 = zero / features unavailable
```

For implementations where `CPUID.7.0:EAX` reports the maximum input subleaf, the second guard would require that value to be at least 1.

### Potential impact

The generic failure mode is a false-positive CPU-feature result. A caller may then dispatch to an instruction that the CPU does not support. The expected failure would be an invalid-instruction exception or equivalent process failure.

I am not claiming a memory-safety vulnerability or assigning a security severity.

### `sha2` context

As one concrete consumer, `sha2` 0.10.9 does not dispatch its x86 SHA-256 backend based only on the `"sha"` bit. It requests the conjunction:

```rust
cpufeatures::new!(shani_cpuid, "sha", "sse2", "ssse3", "sse4.1");
```

Its SHA-512 x86 backend separately requests `"avx2"`.

Therefore practical reachability for that consumer depends on the complete conjunction of required bits, not merely one misinterpreted leaf-7 bit. This conditions or reduces the practical false-positive surface for this specific consumer, but does not correct the crate-wide fail-open query policy.

## 2. AVX uses the XMM-only XCR0 category

This is independent of the maximum-basic-leaf issue above.

### Source evidence

Both 0.2.17 and 0.3.0 define:

```rust
("avx", "xmm", 0, ecx, 28)
```

The `"xmm"` category calls `__xgetbv!` with mask `0b10`, while the existing `"ymm"` category uses `0b110`.

Before reading XCR0, `__xgetbv!` checks CPUID.1:ECX bits 26 and 27, i.e. XSAVE and OSXSAVE. The AVX row also checks CPUID.1:ECX.AVX bit 28. The remaining question is the XCR0 mask.

The usual AVX detection contract is:

```text
CPUID.1:ECX.XSAVE == 1
CPUID.1:ECX.OSXSAVE == 1
XGETBV(XCR0) & 0b110 == 0b110
CPUID.1:ECX.AVX == 1
```

That requires both XMM and YMM state to be enabled by the operating system. Intel's published example likewise checks `XCR0 & 6 == 6`:

https://www.intel.com/content/dam/develop/external/us/en/documents/how-to-detect-new-instruction-support-in-the-4th-generation-intel-core-processor-family.pdf

The current AVX row checks only the XMM bit in XCR0. This appears to permit `"avx"` to report true when XMM state is enabled but YMM state is not.

### Expected behavior / proposed direction

If there is no intentional special case, the AVX mapping appears to need the existing `"ymm"` category:

```rust
("avx", "ymm", 0, ecx, 28)
```

Tests should cover a state with XSAVE, OSXSAVE, and CPUID.AVX set, XMM enabled in XCR0, but YMM disabled, and require AVX detection to return false.

### Potential impact

A false-positive AVX result could allow a consumer to execute AVX instructions without the required OS-managed YMM state, leading to an invalid-instruction fault or equivalent failure.

The current `sha2` 0.10.9 paths inspected here request the SHA/SSE conjunction and AVX2, not the standalone `"avx"` feature. Its AVX2 row already uses the `"ymm"` category. Thus this second finding is crate-wide and independent of the concrete `sha2` instantiation above.

## Limits of what has been demonstrated

For the leaf-7 observation, I have not identified a documented physical platform on which the exact bit conjunction used by the `sha2` consumer above produces a false positive. This report does not claim that such a platform does not exist.

A virtualized environment or hypervisor which caps the visible maximum basic CPUID leaf is a plausible scenario for reproducing the generic fail-open behavior, but this is currently a hypothesis.

For the AVX XCR0 observation, I likewise do not yet have a physical or virtual-machine reproducer demonstrating a field failure. It is a source-level potentially incomplete check presented for maintainer confirmation.

Neither observation is being presented as a demonstrated memory-safety vulnerability, and no security severity is being assigned.

## Questions for maintainers

1. Is the unguarded leaf-7 access based on an intended minimum-CPU or minimum-virtual-CPU assumption? If so, should that assumption be documented?
2. Would you accept a change that obtains `CPUID(0).EAX`, zeros leaf-7 results below the supported maximum, and separately guards subleaf 7.1?
3. For leaf 7.1, which cross-vendor subleaf-availability check do you prefer?
4. Is the `"avx"` → `"xmm"` mapping intentional? If so, what makes the XCR0 YMM-state bit unnecessary for this API's AVX meaning?
5. Would mocked or synthetic CPUID-result tests be acceptable for both boundary conditions?
6. Would you prefer these two independent observations as separate issues?

No patch is attached, and no physical reproducer is currently available.

eslerm:
Closes #1510

**1. Guard CPUID queries by max basic leaf.** `CPUID(1)` and `CPUID(7, N)` were
queried unconditionally. Per the SDM a leaf above `CPUID(0).EAX` returns the
highest supported leaf's data, which `check!` then reads as feature bits. The
leaf 7 rows carrying no XSAVE requirement are where this bites, `sha` and `adx`
among them, since nothing masks a stale bit once it is read.

Leaf 1 and both leaf 7 subleaves are zeroed when `CPUID(0).EAX` does not reach
them. Subleaf 1, which the `sha512`/`sm3`/`sm4` rows read, is additionally gated
on subleaf 0 reporting a max subleaf of at least 1. `std_detect` gates both the
same way:

https://docs.rs/crate/std_detect/latest/source/src/detect/os/x86.rs

**2. Match XSAVE tags to the encodings they gate.** None of the
three is an AVX-512 feature. `ymm` for the VEX-256 forms, none for the
legacy-SSE form of `gfni`, as with `pclmulqdq`. `gfni` needs `""` rather than
`"ymm"`: Tremont Atom has GFNI and no AVX at all. `sm3` had the opposite error,
`"xmm"` where its VEX.128 encoding needs `ymm`; fixed the same way the `avx` row
was in #1511, leaving no `"xmm"` rows. Requires the max-basic-leaf
guard above: without it a part whose max basic leaf is below 7 aliases the leaf
7 query, and the relaxed tags no longer mask the aliased bits.

**3. Make the guard and the tags unit-testable.** Leaf collection moves into
`__collect_leaves`, parameterized over the CPUID sources. No CI runner reports a
max basic leaf below 7, so the guard is otherwise unreachable in tests.
`mod x86` becomes `pub mod x86`, doc-hidden, since the macro now reaches that
function through `$crate::x86::` when expanded downstream; `aarch64` and
`loongarch64` are already public on the same terms.

**4. Require the AVX bit** for `vaes`/`vpclmulqdq`. The `ymm` tag checks that the
OS enabled YMM state, not that the CPU reports AVX; `fma` and `avx2` already
conjoin both. `sha512` and `sm4` are also VEX-encoded and still do not conjoin
it; left alone here, since they were added as a set and are worth their
own change.

**5. Make an unknown XSAVE tag a compile error.** The tag column was matched at
runtime with a `_ => true` arm, so a mistyped tag did not fail to build. It
silently removed the state check for that feature. Feature names were already
protected this way; the tag column was not.

`vaes`/`vpclmulqdq` no longer imply AVX-512 state. A consumer dispatching an
EVEX-512 form must gate on `avx512f` separately.

## Downstream

`aes` gates its VAES-256 backend on the bare `vaes` token, so this changes which
backend it selects on non-AVX-512 parts. That backend batches 30 blocks and does
not override `encrypt_tail_blocks`, so shorter buffers fall through to `cipher`'s
one-block-at-a-time default.

`Aes128::encrypt_blocks`, cycles/byte: below 480 bytes VAES-256 runs up to 34%
slower than AES-NI; at and above it, 4 to 5 times faster. The crossover is
exactly the batch width, and the tail cost recurs at every `N mod 30` (a
45-block buffer is one full batch plus a 15-block tail, and lands at 1.8x rather
than 5x).

Two limits on those numbers. They were taken on Zen 5, which is **not** one of
the parts this change affects; it was the x86 hardware I had, and a Gracemont or
Zen 3 result may differ, particularly the size of the win. And they force backend
selection rather than masking CPUID, so they measure the two backends, not the
detection.

The tail path is a pre-existing `aes` gap. An `encrypt_tail_blocks` override
there would bound it at parity.

## Testing

`cargo test -p cpufeatures` passes on `x86_64-unknown-linux-gnu`; on non-x86
targets the module and its tests are `cfg`'d out. Three of the seven tests fail
against the pre-fix code.

Detection was checked under masked CPUID; no test in this PR produces these.
Rows 1, 3 and 4 are `qemu-x86_64` (qemu-user, TCG) with the named `-cpu` mask;
row 2 is a KVM guest.

| configuration | before | after |
|---|---|---|
| `-cpu Icelake-Server,-avx512f` | `vaes=false` | `vaes=true` |
| KVM `-cpu Snowridge` (GFNI, no AVX) | `gfni=false` | `gfni=true` |
| `-cpu Icelake-Server,level=1` | `vaes=false` | `vaes=false` |
| `-cpu Skylake-Client` | `vaes=false` | `vaes=false` |

The `level=1` row is why the guard comes first. Both shipped states are correct
there; a fourth state this PR never produces, relaxation applied with the guard
absent, reports `vaes=true` on a configuration with no leaf 7 at all.

`vpclmulqdq` was not measured on any configuration, because QEMU's TCG does not
expose its CPUID bit, and no test covers its tag. Its change rests on the same
encoding argument as `vaes`.

### DIFF
--- a/cpufeatures/src/lib.rs
+++ b/cpufeatures/src/lib.rs
@@ -17,7 +17,8 @@ pub mod loongarch64;
 
 #[cfg(not(miri))]
 #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
-mod x86;
+#[doc(hidden)]
+pub mod x86;
 
 #[cfg(miri)]
 mod miri;
--- a/cpufeatures/src/x86.rs
+++ b/cpufeatures/src/x86.rs
@@ -28,42 +28,64 @@ macro_rules! __unless_target_features {
     }};
 }
 
+#[cfg(target_arch = "x86")]
+use core::arch::x86::CpuidResult;
+#[cfg(target_arch = "x86_64")]
+use core::arch::x86_64::CpuidResult;
+
+const ZERO_CPUID: CpuidResult = CpuidResult {
+    eax: 0,
+    ebx: 0,
+    ecx: 0,
+    edx: 0,
+};
+
+/// Collect the CPUID leaves used for feature detection: leaf 1, and leaf 7
+/// subleaves 0 and 1.
+///
+/// Querying a leaf beyond the CPU's maximum supported basic leaf does not
+/// return zero: the CPU instead re-returns the highest supported leaf's data
+/// (Intel SDM Vol. 2A, "CPUID"), which `check!` would otherwise misinterpret
+/// as feature bits. The CPUID sources are parameterized so this guard can be
+/// exercised with synthetic data in tests.
+#[doc(hidden)]
+pub fn __collect_leaves(
+    cpuid: impl Fn(u32) -> CpuidResult,
+    cpuid_count: impl Fn(u32, u32) -> CpuidResult,
+) -> [CpuidResult; 3] {
+    let max_leaf = cpuid(0).eax;
+
+    let leaf1 = if max_leaf >= 1 { cpuid(1) } else { ZERO_CPUID };
+    let leaf7_0 = if max_leaf >= 7 {
+        cpuid_count(7, 0)
+    } else {
+        ZERO_CPUID
+    };
+    // `max_leaf >= 7` is redundant here (leaf7_0 is zeroed otherwise, so its
+    // eax cannot reach 1) but is kept to state the precondition explicitly.
+    let leaf7_1 = if max_leaf >= 7 && leaf7_0.eax >= 1 {
+        cpuid_count(7, 1)
+    } else {
+        ZERO_CPUID
+    };
+
+    [leaf1, leaf7_0, leaf7_1]
+}
+
 /// Use CPUID to detect the presence of all supplied target features.
 #[macro_export]
 #[doc(hidden)]
 macro_rules! __detect_target_features {
     ($($tf:tt),+) => {{
         #[cfg(target_arch = "x86")]
-        use core::arch::x86::{__cpuid, __cpuid_count, CpuidResult};
+        use core::arch::x86::{__cpuid, __cpuid_count};
         #[cfg(target_arch = "x86_64")]
-        use core::arch::x86_64::{__cpuid, __cpuid_count, CpuidResult};
+        use core::arch::x86_64::{__cpuid, __cpuid_count};
 
-        unsafe fn cpuid(leaf: u32) -> CpuidResult {
-            __cpuid(leaf)
-        }
-
-        unsafe fn cpuid_count(leaf: u32, sub_leaf: u32) -> CpuidResult {
-            __cpuid_count(leaf, sub_leaf)
-        }
-
-        const ZERO: CpuidResult = CpuidResult { eax: 0, ebx: 0, ecx: 0, edx: 0 };
-
-        let cr = unsafe {
-            // Querying a leaf beyond the CPU's maximum supported basic leaf
-            // does not return zero: the CPU instead re-returns the highest
-            // supported leaf's data (Intel SDM Vol. 2A, "CPUID"), which
-            // `check!` would otherwise misinterpret as feature bits.
-            let max_leaf = cpuid(0).eax;
-            let leaf1 = if max_leaf >= 1 { cpuid(1) } else { ZERO };
-            let leaf7_0 = if max_leaf >= 7 { cpuid_count(7, 0) } else { ZERO };
-            let leaf7_1 = if max_leaf >= 7 && leaf7_0.eax >= 1 {
-                cpuid_count(7, 1)
-            } else {
-                ZERO
-            };
-
-            [leaf1, leaf7_0, leaf7_1]
-        };
+        let cr = $crate::x86::__collect_leaves(
+            |leaf| unsafe { __cpuid(leaf) },
+            |leaf, sub_leaf| unsafe { __cpuid_count(leaf, sub_leaf) },
+        );
 
         $($crate::check!(cr, $tf) & )+ true
     }};
@@ -83,6 +105,10 @@ macro_rules! __xgetbv {
         let xmask = 0b11 << 26;
         let xsave = $cr[0].ecx & xmask == xmask;
         if xsave {
+            // SAFETY: `xsave` above confirms CPUID.1:ECX bits 26 and 27
+            // (XSAVE and OSXSAVE) are both set, so XGETBV is supported by
+            // the CPU and enabled by the OS. ECX = 0 is always a valid
+            // XCR index.
             let xcr0 = unsafe { arch::_xgetbv(arch::_XCR_XFEATURE_ENABLED_MASK) };
             (xcr0 & $mask) == $mask
         } else {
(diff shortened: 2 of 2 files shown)


eslerm:
Closing for now — opened prematurely, will reopen when ready.

Issue #1472, opened 2026-04-19
aarch64-dit crate is missing fencing (for both compiler and CPU)


thomcc:
The `aarch64-dit` crate has two notable flaws:
1. It is possible for the compiler to reorder operations outside the critical section where DIT is enabled.
2. It does not perform speculation fencing, which is documented as necessary at least on apple platforms in this document: https://developer.apple.com/documentation/xcode/writing-arm64-code-for-apple-platforms#Enable-DIT-for-constant-time-cryptographic-operations.

# Flaw 1: Compiler Reordering

Here's an example of the first flaw: https://godbolt.org/z/8fdsnxrc5. Essentially the compiler is free to move instructions outside of the `asm!("msr DIT #1")` and `asm!("msr DIT #0")` guarded region, and you can see that it essentially compiles:
```rs
if have_dit {
    dit_enable();
}
let res = sensitive(a, b);
if have_dit {
    dit_disable();
}
return res;
```
into
```rs
if have_dit {
    dit_enable();
    dit_disable();
}
return sensitive(a, b);
```
which defeats the point, since the instructions which are data-dependent are no longer in the block where DIT is enabled.

In an ideal world, inserting compiler fences (or atomic fences, or adjusting the `asm!` options...) would fix this, however they only apply to memory accesses, so that does not work. The only approach I can think of that does work is to use an indirect jump whose target we obscure from the vision of the compiler. Something like:

```rs
impl Dit {
    // ...snip ...
    /// Call `f` with DIT enabled, returning the result. Takes care to ensure
    /// that that any computations which occur during `f` do not leak outside
    /// the critical section.
    pub fn with<R, F: FnOnce() -> R>(s: &self, f: F) -> R {
        let guard = self.enable();
        let mut o: Option<F> = Some(f);
        // Note: Could be unwrap_unchecked().
        let f: &mut dyn FnMut() -> R = &mut || o.take().unwrap()();
        core::hint::black_box(f)()
    }
    // ...snip...
}
```

This works, as demonstrated here: https://godbolt.org/z/fv1Gofjxc (note: the `add` instruction was changed to floating point addition just so that it's clearer in the assembly listing what is happening—you can see that the `fadd` takes place in the indirectly called function, which is called (the `blr x8`) between the `msr DIT #1;` and `msr DIT #0` instructions.

# Flaw 2: Speculation

[...]

PR #1536, opened 2026-09-10
cpufeatures: add Speculation Barrier (SB) support for AArch64

tarcieri:
As suggested in #1472, this instruction is needed by the `aarch64-dit` crate to properly implement Data-Independent Timing:

https://support.arm.com/documentation/ddi0487/mb/-Part-B-The-AArch64-Application-Level-Architecture/-Chapter-B2-The-AArch64-Application-Level-Memory-Model/-B2-6-Memory-barriers/-B2-6-3-Speculation-Barrier

The usage on Apple targets is covered here:

https://developer.apple.com/documentation/xcode/writing-arm64-code-for-apple-platforms#Enable-DIT-for-constant-time-cryptographic-operations

> "To ensure that subsequent instruction timing reflects the updated DIT
> processor state, add a speculation barrier after turning on DIT.
> Use the sb instruction to add a speculation barrier"

This PR is effectively an extraction of the `cpufeatures`-related changes originally suggested in this patch from #1472:

https://github.com/RustCrypto/utils/commit/f31f41cb4c5089b7ce0fd20b818f335a64bbf2c3

### DIFF
--- a/cpufeatures/src/aarch64.rs
+++ b/cpufeatures/src/aarch64.rs
@@ -67,10 +67,11 @@ macro_rules! __expand_check_macro {
 #[cfg(any(target_os = "linux", target_os = "android"))]
 __expand_check_macro! {
     ("aes",    AES),    // Enable AES support.
-    ("dit",    DIT),    // Enable DIT support.
+    ("dit",    DIT),    // Enable Data-Independent Timing support.
     ("sha2",   SHA2),   // Enable SHA1 and SHA256 support.
     ("sha3",   SHA3),   // Enable SHA512 and SHA3 support.
     ("sm4",    SM4),    // Enable SM3 and SM4 support.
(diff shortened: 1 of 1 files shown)


cpufeatures/src/aarch64.rs
```diff
@@ -89,6 +90,7 @@ pub mod hwcaps {
     pub const SHA2: c_ulong = libc::HWCAP_SHA2;
     pub const SHA3: c_ulong = libc::HWCAP_SHA3 | libc::HWCAP_SHA512;
     pub const SM4: c_ulong = libc::HWCAP_SM3 | libc::HWCAP_SM4;
+    pub const SB: c_ulong = libc::HWCAP_SB;
```
tarcieri, line 93: Note: it's unclear to what extent either DIT or SB are helpful on non-Apple targets, as covered in #1472, but they've been added here (i.e. Linux) for completeness

PR #1543, opened 2026-09-18
Allow triggering CI workflows manually

nazar-pc:
This helps contributors to check their changes by triggering workflow manually from GitHub Actions page of their fork before opening PR and annoying everyone with notifications


newpavlov:
I don't think I've seen this option used in other high-profile Rust projects. If I need to run CI for development purposes (e.g. if project requires involved setup procedure), I usually just open PR against my fork.

nazar-pc:
> I don't think I've seen this option used in other high-profile Rust projects

`syn` with 2.4B downloads [has it](https://github.com/dtolnay/syn/blob/2ebf5f96e7361e50bfc3fa28d77908c445a34c4f/.github/workflows/ci.yml#L6), `libc` with another 1.6B downloads on crates.io [has it too](https://github.com/rust-lang/libc/blob/35c8b7d847f64265cb578343dc5d1be467fe63db/.github/workflows/ci.yaml#L12), and many more.

---

If you're suggesting that this shouldn't be added because many developers apparently do not know about it then I'd say it is a weak argument and adding it cost basically nothing. It is also beneficial for upstream developers: you can test changes in any branch without opening PRs too.

The alternative of opening PR just to run CI works, but is essentially a noisy workaround that you shouldn't need in the first place. It is also mostly done against upstream and annoys maintainers with countless useless GitHub notifications.

Another alternative is to trigger CI on every single push to any branch, which clogs CI queue for no good reason most of the time.

newpavlov:
>syn with 2.4B downloads [has it](https://github.com/dtolnay/syn/blob/2ebf5f96e7361e50bfc3fa28d77908c445a34c4f/.github/workflows/ci.yml#L6), libc with another 1.6B downloads on crates.io [has it too](https://github.com/rust-lang/libc/blob/35c8b7d847f64265cb578343dc5d1be467fe63db/.github/workflows/ci.yaml#L12), and many more.

Ah, good to know! In that case I am fine with merging this. It's also worth to do this in other repos as well for consistency.

